PR #1, opened 2016-03-08
Fake PR just to re-start the CIs

ocefpaf:
Don't need to merge as this should re-start the CIs and get the files uploaded.


ocefpaf:
PY35 failed as expected. See https://github.com/conda-forge/staged-recipes/pull/49#issuecomment-193835732

Issue #10, opened 2016-03-16
gdal2 branch windows packages not uploaded


kmuehlbauer:
If this is the same as in conda-forge/wradlib-feedstock#5 I could issue the necessary PR.

pelson:
Just needs #4 to be merged I think.

kmuehlbauer:
@pelson Is there any guideline within conda-forge, when to merge such adminstrative stuff. Also the mentioned pr is directed into master branch not gdal2 branch. Wow do we deal with it? Merge into master and then add then apply the changes also to gdal2 branch?

pelson:
> Also the mentioned pr is directed into master branch not gdal2 branch.

Aha! Yes, I'd forgotten. This was/is the first feedstock to have multiple branches - some of the administrative stuff won't handle multiple branches yet, I need to catch up on those cases. In the meantime, yes, I think it is wise to apply the changes from #4 into the gdal2 branch - can this be achieved cleanly by merging master into the gdal2 branch?

kmuehlbauer:
If the two branches diverge more and more, it can be very problematic to apply automatic changes. I have just tried to create a PR from conda-forge-admin/gdal-feedstock:feedstock-renderer directly to conda-forge/gdal-feedstock:gdal2. This would work. 

If  #4 is merged before, we could try to do this pr from master to gdal2 branch. If this doesn't work out, then we can merge from conda-forge-admin. Can anybody follow?

pelson:
Yep. Sounds sensible to me @kmuehlbauer. You have merge rights on this repo, so give it a shot! :+1:

kmuehlbauer:
@pelson To get this right, we merge #4 now and then apply the changes to gdal2-branch.

pelson:
Yep. 

> apply the changes to gdal2-branch.

As in:

```
git checkout -b gdal2 upstream/gdal2
git merge upstream/master
... check as expected
git push origin gdal2_pr
... create a PR which proposes the merge to gdal2 from your gdal2_pr branch
```

kmuehlbauer:
@pelson Now we have this pr here #11. I hope, it's OK that way.

pelson:
LGTM. Keep an eye on conda-forge channel on anaconda.org, and I suspect we can close this with no further work. :+1:

kmuehlbauer:
FYI: The four windows python27 versions are uploaded, ... finally.

pelson:
[...]

PR #11, opened 2016-03-16
MNT: update binstar tokens

kmuehlbauer:


pelson:
:+1: - one thing to watch out for - it is safer to push your branch to your fork, rather than the forge feedstock as any branches on conda-forge will be built and uploaded to anaconda.org/conda-forge.
Otherwise, thanks @kmuehlbauer!

kmuehlbauer:
@pelson I thought about that twice before pushing this to conda-forge, went too strict along your [comments](https://github.com/conda-forge/gdal-feedstock/issues/10#issuecomment-197215873). Got `origin` in the wrong context. Will have pr generate from my fork in the future.

PR #13, opened 2016-03-26
features

ocefpaf:
Note to self: If this works here do the same for `gdal2` branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ env:
 install:
     - |
       MINICONDA_URL="http://repo.continuum.io/miniconda"
-      MINICONDA_FILE="Miniconda3-3.7.3-MacOSX-x86_64.sh"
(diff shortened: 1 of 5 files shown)


pelson:
I tried implementing multibranch MNT PRs on this feedstock at #14 and #15. Happy for them to be ignored, but I'm getting close to having the necessary tools for multibranch support. :+1:

ocefpaf:
Ah got it now.

ocefpaf:
The thing is that #13 also add features, which is not handled automatically, right?

pelson:
> The thing is that #13 also add features, which is not handled automatically, right?

Right. The manual approach is the only way when you want to make functional changes (I can't script that bit out :wink:)

ocefpaf:
> Right. The manual approach is the only way when you want to make functional changes (I can't script that bit out :wink:)

:relieved: I am not completely obsolete!

PR #17, opened 2016-03-27
Link against existing Conda libs where possible and fix GDAL_DATA env value

gillins:
This PR links against Conda provided libs for tiff, PNG, jpeg and zlib. This saves space and reduces likelihood of conflicts in other libs.

This wasn't possible zlib and jpeg under Windows since the GDAL build doesn't have an option for an external zlib and Conda jpeg isn't provided as a DLL.

Under Windows, this PR also fixes the syntax problem in activate.bat - %GDAL_DATA% ended up with quotes in it with GDAL doesn't handle. HDF4 support is removed since the Conda package doesn't yet support VC features and will be linked against wrong MS runtimes for Python 3.x.


recipe/bld.bat
```diff
@@ -21,13 +21,17 @@ set BLD_OPTS=%WIN64% ^
     XERCES_DIR=%LIBRARY_PREFIX% ^
     XERCES_INCLUDE="-I%LIBRARY_INC% -I%LIBRARY_INC%\xercesc" ^
     XERCES_LIB=%LIBRARY_LIB%\xerces-c_3.lib ^
-    HDF4_DIR=%LIBRARY_PREFIX% ^
-    HDF4_LIB="%LIBRARY_LIB%\hdf.lib %LIBRARY_LIB%\mfhdf.lib %LIBRARY_LIB%\xdr.lib" ^
```
ocefpaf: Why did you remove the HDF4 on Windows? Was that a problem?

recipe/build.sh
```diff
@@ -25,15 +29,6 @@ CPPFLAGS="-I$PREFIX/include" LDFLAGS="-L$PREFIX/lib" \
...
 
-if [ $(uname) == Darwin ]; then
-    # Copy TIFF and GeoTIFF Headers to build against internal libraries.
-    mkdir -p $PREFIX/include/gdal/frmts/gtiff/libgeotiff
-    cp frmts/gtiff/libgeotiff/*.h $PREFIX/include/gdal/frmts/gtiff/libgeotiff
-    cp frmts/gtiff/libgeotiff/*.inc $PREFIX/include/gdal/frmts/gtiff/libgeotiff
-    mkdir -p $PREFIX/include/gdal/frmts/gtiff/libtiff
-    cp frmts/gtiff/libtiff/*.h $PREFIX/include/gdal/frmts/gtiff/libtiff
-fi
-
```
ocefpaf: :+1: 
Can we add a test to check if the tiff functionalities of `gdal` are there?

recipe/meta.yaml
```diff
@@ -42,6 +46,10 @@ requirements:
         - libnetcdf
         - kealib
         - krb5
+        - libtiff
+        - libpng
+        - zlib  # [not win]
+        - jpeg  # [not win]
```
ocefpaf: I am not sure but I guess that the default channel has `zlib` and `jpeg` for Windows, no?

recipe/meta.yaml
```diff
@@ -85,3 +93,4 @@ extra:
   recipe-maintainers:
     - ocefpaf
     - kmuehlbauer
+    - gillins
```
ocefpaf: :+1:

recipe/scripts/activate.bat
```diff
@@ -1,4 +1,4 @@
 if not defined GDAL_DATA (
-  set GDAL_DATA="%~dp0\..\..\..\Library\share\gdal"
+  set "GDAL_DATA=%~dp0\..\..\..\Library\share\gdal"
```
ocefpaf: Ah! Thanks!!

@bekozi with this fix we should go back and try ocgis again.

ocefpaf:
I'd love to get @kmuehlbauer, @jakirkham, and @groutr opinions here

recipe/bld.bat
gillins: Yes - HDF4 on Windows is only available as a 'pre-VC Features' build - hence built with the wrong compiler for Python 3.x. HDF4 needs to be brought up to date (maybe another conda-forge project?). With the wrong compiler random application crashes will be likely.

recipe/meta.yaml
gillins: Yes - but under Windows GDAL cannot be build against an external zlib for some reason (there is no option to do this). jpeg isn't available as a DLL under Windows, and the provided .lib is built for use in a .exe rather than a .dll so it won't even link to GDAL (another conda-forge project??)

jakirkham:
What sort of feedback are you looking for? I don't really use GDAL so don't have a great idea of what works well or not with this package. My understanding is @gillins is trying to make sure we use conda packages for the various image libraries instead of possibly using system libraries. That sounds good to me. If there are particular details that you would like me to look at, could you please draw my attention to them, @ocefpaf?

recipe/scripts/activate.bat
gillins: Note that setting of environment variables from activate.bat may still be broken on Windows by default - see https://github.com/conda/conda-env/issues/218 and linked PR. If you copy the changes over manually (https://github.com/conda/conda-env/pull/219) it seems to work fine.

recipe/bld.bat
ocefpaf: Thanks for the explanation. I believe that `hdf4` is [*under construction*](https://github.com/conda-forge/staged-recipes/pull/122).

Let's open an issue for that.

recipe/meta.yaml
ocefpaf: Thanks again! (And again I am opening issues to keep track of this.)

ocefpaf:
> What sort of feedback are you looking for?

Basic input from gdal power users.

>  I don't really use GDAL 

So we are on the same page :smile:, but I believe @groutr has more experience. Merging this anyway as it looks good to me. We can iterate on the new issues later.

Thanks @gillins!

recipe/scripts/activate.bat
ocefpaf: Thanks!

kmuehlbauer:
@ocefpaf Sorry, didn't had time to have a closer look. As I'm using linux only, I can't give feedback on the windows issue.

+1 for using conda packages wherever available.

PR #32, opened 2016-04-07
Re-render and add libzma

ocefpaf:
I had problems compiling `libxml2` with `libzma` support, but after removing `libzma` from `libxml2` both `gdal` and `fiona` had issues with missing `libzma`. Let's try this approach for now, but I suspect my original problem was because `homebrew` was not properly removed (my bad [here](https://github.com/conda-forge/conda-smithy/pull/114/files#diff-4db594a93663b6584405af3d0607f2b6R30)). So I will re-visit this at a later time and maybe re-add `libzma` on `libxml2`. For now this seems to work and we need to get `gdal` and `fiona` out there.


ocefpaf:
I had to remove `libspatialite` (`libxml2`) on OS X because of:

```
install_name_tool -add_rpath @loader_path/../lib/ /Users/travis/miniconda3/envs/_build/bin/gdaladdo
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: changing install names or rpaths can't be redone for: /Users/travis/miniconda3/envs/_build/bin/gdaladdo (for architecture x86_64) because larger updated load commands do not fit (the program must be relinked, and you may need to use -headerpad or -headerpad_max_install_names)
Error: install_name_tool failed with exit status 1
```

PS: I tried both `-headerpad or -headerpad_max_install_names` and neither worked.

PR #43, opened 2016-04-25
Rebuild

ocefpaf:
Should enable the missing features on `Python 3.4` and `3.5`.

PR #44, opened 2016-04-26
Bump build number

ocefpaf:
Forgot to do this in #43. (I guess it is time to call it a day and :zzz:)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: objective-c
 
 env:
   matrix:
-
+    
     - CONDA_NPY=110  CONDA_PY=27
     - CONDA_NPY=111  CONDA_PY=27
     - CONDA_NPY=110  CONDA_PY=34
@@ -33,7 +33,7 @@ install:
       conda update --yes conda
       conda install --yes conda-build=1.20.0 jinja2 anaconda-client
       conda config --add channels conda-forge
-
+      
 
 script:
   - conda build ./recipe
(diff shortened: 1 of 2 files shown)


gillins:
@ocefpaf what is needed to start new builds of GDAL 2? https://github.com/conda-forge/gdal-feedstock/pull/42 has merged but files haven't appeared on anaconda.org...

ocefpaf:
@gillins you need to bump the build number. Sorry for missing that in your PR. (I was away with a phone and it was hard to read everything.)

PS: I believe we can pin to `jpeg 9b` now. The whole stack seems to the in line with it.

gillins:
Ah. Can't believe I forgot that. I'll make a new PR to pin to 9* and bump the version number.

ocefpaf:
> Ah. Can't believe I forgot that. I'll make a new PR to pin to 9* and bump the build number.

No problem. In fact that was better since now that I (believe) fixed the jpeg pinning. See if you can aling your PR with the changes in https://github.com/conda-forge/gdal-feedstock/pull/43/files for the `build.sh` and `meta.yaml` files.

gillins:
Why are you excluding some drivers for Python 3?

ocefpaf:
> Why are you excluding some drivers for Python 3?

Why am I putting them back :wink:

I downgraded `xz` to `5.0.5` because `xz 5.2.2` did not work with Python 3.4/3.5 so now I can put anu dependency on `xz` back.

PR #45, opened 2016-04-26
Pin GDAL 2.0 to jpeg 9* and enable external jpeg and hdf4 for Windows

gillins:


gillins:
Probably a good idea if https://github.com/conda-forge/staged-recipes/pull/442 is merged before this one as it changes the version of the .so/.dylib that GDAL links against.

ocefpaf:
@gillins this will be compiled against the new `kealib`. Thanks for both PRs!

gillins:
Thanks - seems to be working well for me now

PR #47, opened 2016-04-27
Enable PAM

### LINKED ISSUE #46: ISIS3/USGS driver problem
This version (or maybe only this compile) suffers from a recently popping up problem with USGS ISIS3 cube files. It was previously seen years ago, when it was suggested that it can happen from an unclean compile, see here for a discussion:
https://lists.osgeo.org/pipermail/gdal-dev/2009-September/022101.html

Symptom: Opening an USGS ISIS cube errors with:
```bash
ERROR 6: SetNoDataValue() not supported for this dataset.
ERROR 6: SetOffset() not supported on this raster band.
ERROR 6: SetScale() not supported on this raster band.
```
while this driver is well supported for many years.

I just tried the recently updated pure conda version of gdal, and that one currently does not have this problem.
Here is a mini cube file to try things out: https://dl.dropboxusercontent.com/u/139035/cropped.cub
When called with this version of gdal it looks like this:
```bash
(gdal_forge)└─❱❱❱ gdalinfo cropped.cub                                                       +5067 18:11 ❰─┘
ERROR 6: SetNoDataValue() not supported for this dataset.
ERROR 6: SetOffset() not supported on this raster band.
ERROR 6: SetScale() not supported on this raster band.
Driver: ISIS3/USGS Astrogeology ISIS cube (Version 3)
Files: cropped.cub
Size is 100, 100
Coordinate System is:
PROJCS["Sinusoidal Mars",
    GEOGCS["GCS_Mars",
        DATUM["D_Mars",
            SPHEROID["Mars",3396190,0]],
        PRIMEM["Reference_Meridian",0],
        UNIT["degree",0.0174532925199433]],
    PROJECTION["Sinusoidal"],
    PARAMETER["longitude_of_center",296.28065864847],
    PARAMETER["false_easting",0],
    PARAMETER["false_northing",0]]
Origin = (-4959.626702915599708,-4803213.120008500292897)
Pixel Size = (0.495566217317710,-0.495566217317710)
Corner Coordinates:
Upper Left  (   -4959.627,-4803213.120) ( 64d15'22.21"W, 81d 1'59.20"S)
Lower Left  (   -4959.627,-4803262.677) ( 64d15'22.39"W, 81d 2' 2.21"S)
Upper Right (   -4910.070,-4803213.120) ( 64d15' 2.90"W, 81d 1'59.20"S)
Lower Right (   -4910.070,-4803262.677) ( 64d15' 3.07"W, 81d 2' 2.21"S)
Center      (   -4934.848,-4803237.898) ( 64d15'12.64"W, 81d 2' 0.70"S)
Band 1 Block=100x100 Type=Float32, ColorInterp=Undefined
```

When called with conda's gdal, it does not show this error:
```bash
(gdal_try)└─❱❱❱ gdalinfo cropped.cub                                                         +5070 18:13 ❰─┘
Driver: ISIS3/USGS Astrogeology ISIS cube (Version 3)
Files: cropped.cub
Size is 100, 100
Coordinate System is:
PROJCS["Sinusoidal Mars",
    GEOGCS["GCS_Mars",
        DATUM["D_Mars",
            SPHEROID["Mars",3396190,0]],
        PRIMEM["Reference_Meridian",0],
        UNIT["degree",0.0174532925199433]],
    PROJECTION["Sinusoidal"],
    PARAMETER["longitude_of_center",296.28065864847],
    PARAMETER["false_easting",0],
    PARAMETER["false_northing",0]]
Origin = (-4959.626702915599708,-4803213.120008500292897)
Pixel Size = (0.495566217317710,-0.495566217317710)
Corner Coordinates:
Upper Left  (   -4959.627,-4803213.120) ( 64d15'22.21"W, 81d 1'59.20"S)
Lower Left  (   -4959.627,-4803262.677) ( 64d15'22.39"W, 81d 2' 2.21"S)
Upper Right (   -4910.070,-4803213.120) ( 64d15' 2.90"W, 81d 1'59.20"S)
Lower Right (   -4910.070,-4803262.677) ( 64d15' 3.07"W, 81d 2' 2.21"S)
Center      (   -4934.848,-4803237.898) ( 64d15'12.64"W, 81d 2' 0.70"S)
Band 1 Block=100x100 Type=Float32, ColorInterp=Undefined
  NoData Value=-3.40282265508890445e+38
```

ocefpaf:
I suspect this might be the cause of #46. If that is true it is unfortunate that the change in the internal `giflib` never made it upstream and that an external source is even allowed when compiling `gdal`.

@gillins I am going to :zzz: if this is green can you merge and propagate the change to `gdal 2`?

Fixes #46


recipe/meta.yaml
```diff
@@ -78,6 +76,7 @@ test:
         - sites.sbx
         - sites.shp
         - sites.shx
+        - cropped.cub
```
ocefpaf: @michaelaye I am using your sample file in the test to avoid regression. Is that OK with you?

recipe/run_test.sh
```diff
@@ -16,3 +16,6 @@ echo "odata_value -9999" >> grid.hdr
 proj4="+y_0=2400761.714982585 +lat_ts=-19.6097 +a=6378137.0 +proj=merc +units=m +b=6356752.3142 +lat_0=-19.6097 +x_0=-0.0 +lon_0=-70.7691"
 
 gdalwarp -s_srs "+proj=latlong" -t_srs "$proj4" -of EHdr grid.asc grid.flt
+
+# Test ISIS3/USGS driver `SetNoDataValue()` issue.
+gdalinfo cropped.cub
```
ocefpaf: This does not return a non-zero status. I need to think more about this test.

ocefpaf:
Closing due to https://github.com/conda-forge/gdal-feedstock/pull/47/files#r61190447

I will take a better look at this tomorrow.

recipe/meta.yaml
michaelaye: That's fine, thanks for debugging this! All planetary python it's will be thankful if we can haz a stable GDAL package!

ocefpaf:
From @ceholden in https://github.com/conda-forge/gdal-feedstock/issues/46#issuecomment-214978823

> Aha! I tracked down the issue to PAM being disabled as well and can confirm rebuilding without PAM disabled fixes it.

I can confirm that from a local build. Now I am testing it here.

> When I enable PAM using `GDAL_PAM_ENABLED=YES gdalinfo <file>`, the PDS files are read correctly and this change didn't cause new problems in any of the other formats (i.e., no regressions by turning PAM on).

That seems to be a :+1: to enable `PAM`

> I don't know if this is the correct backstory, but if the recipe was ported from the [IOOS channel](https://github.com/ioos/conda-recipes) then the `--without-pam` has been around in the Unix build [since the first commit of the recipe](https://github.com/ioos/conda-recipes/commit/f2a98411ddf1604e82adf0a50a7a8690fb45fe8b).

IOOS `gdal` was based on OSGEO's `gdal` and they disable `PAM`. Not sure why. (That is +0 for enabling PAM :wink:)

> There is [some discussion on the GDAL mailing list from 2007 when they changed PAM to be enabled by default that lacks objections](https://lists.osgeo.org/pipermail/gdal-dev/2007-September/014159.html). I know some of my peers tend to delete these metadata to reduce the number of files on disk (which can make an impact when you dump 1000s of images into one directory).

Is that still true?

> There are a few other motives for disabling PAM I've found, including [a personal preference against the PAM files when batch processing](https://lists.osgeo.org/pipermail/gdal-dev/2009-January/019580.html) and [GDAL reading out of date raster statistics from PAM instead of recomputing](https://lists.osgeo.org/pipermail/gdal-dev/2009-January/019580.html).

That is a :-1: for enabling `PAM`.

> The last reason given has tripped me up before, but disabling PAM really clobbers the PDS formats since I'm not sure you can even open the datasets in Python. As such, I would be inclined to enable it barring other objections:

Thanks for your assessment. I am not a `gdal` expert, but it seems that enabeling/disableing at build time does not remove any functionality as we ca sett an env var (`GDAL_PAM_ENABLED`) and get it back.

The question gets easier than: Who should we follow? The default channel or the OSGEO channel? Which option is better to be the default one?

``` bash
> python -c 'from osgeo import gdal; ds = gdal.Open("cropped.cub"); print(ds)'
ERROR 6: SetNoDataValue() not supported for this dataset.
ERROR 6: SetOffset() not supported on this raster band.
ERROR 6: SetScale() not supported on this raster band.
None
> GDAL_PAM_ENABLED=YES python -c 'from osgeo import gdal; ds = gdal.Open("cropped.cub"); print(ds)'
<osgeo.gdal.Dataset; proxy of <Swig Object of type 'GDALDatasetShadow *' at 0x7f48c8f4e5d0> >
```

@ceholden I am assuming you are :+1: to enable `PAM`.
@gillins what is your opinion?
I am +0.5 right now, but I would like to at least get our Windows builds in par with the *nix builds.

gillins:
I think I'm +0.5 also. Matching the default channel and GDAL default behaviour makes the most sense but those .xml files are a real pain so I will probably be setting GDAL_PAM_ENABLED=YES straight away on my machine...

ocefpaf:
@msarahan do you know why the default channel chose to enable `PAM`?

(See https://github.com/conda-forge/gdal-feedstock/pull/47#issuecomment-215040770 for more details)

ceholden:
I'm +0.5 as well. I didn't actually know you could enable/disable it before looking into this issue so it's hard to have a strong opinion. The "fewer files on disk" argument I mentioned is a pretty specific and silly reason. It could be better addressed with some directory structuring (e.g., group MODIS data into folders by year, by product, etc. rather than dumping it all in one spot) or, evidently, by turning off PAM manually.

I'm mostly in favor for reasons of consistency with other distributions of GDAL. A few other data points on this:
* I don't claim to understand Debian/Ubuntu packaging, but from [what I could find they don't disable PAM](https://anonscm.debian.org/cgit/pkg-grass/gdal.git/tree/debian/rules?h=ubuntu-trusty&id=b487d03d32cb25e3de23614fe47b931badbee160#n86)
* Arch Linux's GDAL [doesn't disable GDAL](https://projects.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/gdal#n47)
* The Enterprise Linux GIS RPM packages (for CentOS/Redhat) don't disable PAM [in the SRPM](http://elgis.argeo.org/repos/6/elgis/SRPMS/). Not sure if there's a better source for this -- I'm stuck compiling GDAL manually on our CentOS cluster :frowning: 

Is there a way to document these sorts of build decisions inside or alongside the package?

michaelaye:
You guys seem to only talk symptoms though. What about looking at the drivers themselves? Why does the ISIS3 driver only work with PAM enabled? Maybe this could be changed? @thare?

michaelaye:
Sorry I meant to notify @thareUSGS and am on iOS where I can't edit the previous comment.

ocefpaf:
@ceholden just to add two more data points. [Fedora](http://koji.fedoraproject.org/koji/buildinfo?buildID=755897) and [OpenSUSE](https://build.opensuse.org/package/view_file/openSUSE:Factory/gdal/gdal.spec?expand=1) do not disable PAM either

> You guys seem to only talk symptoms though. 

Not at all. We are looking from a packaging point of view. It is just a decision about defaults, since the behavior can easily be turn ON/OFF afterwards with the `GDAL_PAM_ENABLED`. However, feel free to take any question regarding the drivers to the `gdal` devs and report back if you are find anything interesting.

ocefpaf:
All those +0.5 and the fact that the default channel and most Linux distros do not disable `PAM` I am merging this. If anyone finds that this is a problem please report back here.

> Is there a way to document these sorts of build decisions inside or alongside the package?

This PR is the tracker. I am adding a [comment](https://github.com/conda-forge/gdal-feedstock/pull/47/commits/ce90a4cb7a89e6e543eaf25402846195193bc8db) in the recipe with the link

michaelaye:
One remaining confusion for me:

> When I enable PAM using GDAL_PAM_ENABLED=YES gdalinfo <file>, the PDS files are read correctly and this change didn't cause new problems in any of the other formats (i.e., no regressions by turning PAM on).

You guys are aware that the PDS reader and the ISIS3 reader are two different things, right? Because #46 related to ISIS3 only. And the above quoted text talks about PDS files working fine? Or was that only a test to avoid regression on supporting PDS file reading? Obviously one shouldn't break PDS support by fixing ISIS3 support. Thanks so much for debugging this, very happy end user here!

ceholden:
Ah, my bad. I might have gone too far with the statement: "...  ISIS3 is part of a family of related formats including PDS and ISIS2." from [the GDAL driver documentation page for ISIS3](http://www.gdal.org/frmt_isis3.html).

Disabling PAM causes issues for all members of this family of drivers tested -- PDS and ISIS3.

See the GDAL docs for:
* [PDS](http://www.gdal.org/frmt_pds.html)
* [ISIS2](http://www.gdal.org/frmt_isis2.html)
* [ISIS3](http://www.gdal.org/frmt_isis3.html)

Hope this helps! It's been really interesting for me to run into coordinate reference systems for Mars!

PR #48, opened 2016-05-02
Update gdal2 to version 2.1.0

kmuehlbauer:
gdal2 version 2.1.0 is released

- update to gdal version 2.1.0
- reset version


kmuehlbauer:
travis-ci: 
```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated
```

Should we silence `make` or is there a better solution?

kmuehlbauer:
Appveyor: 
```bash
nitfwritejpeg_12.obj : error LNK2019: unresolved external symbol "void __cdecl jpeg_vsiio_dest_12(struct 
jpeg_compress_struct12 *,struct _iobuf *)" (?jpeg_vsiio_dest_12@@YAXPAUjpeg_compress_struct12@@PAU_iobuf@@@Z) 
referenced in function "int __cdecl NITFWriteJPEGBlock_12(class GDALDataset *,struct _iobuf
 *,int,int,int,int,int,int,unsigned char const *,int,int (__stdcall*)(double,char const *,void *),void *)"
 (?NITFWriteJPEGBlock_12@@YAHPAVGDALDataset@@PAU_iobuf@@HHHHHHPBEHP6GHNPBDPAX@Z4@Z)
```


References:
https://lists.osgeo.org/pipermail/gdal-dev/2016-January/043439.html

ocefpaf:
@kmuehlbauer we should probably silence it.

BTW let's take this opportunity to flip things here. I will start a `gdal 1.x` branch and `gdal 2` should go to master. Also, let's try to clean this recipe a little bit by putting those test data files in a directory. (I can do those later if you don't want to do it now.)

PS: I have no idea how to fix the AppVeyor failure.

gillins:
Looks like we might have to disable JPEG12_SUPPORTED in nmake.opt as mentioned in that post. I'll create a patch for this tomorrow.

kmuehlbauer:
@ocefpaf You are referring to the `sites.xyz` files? Or the os1_hw.py and run_test.py also?

ocefpaf:
All data files but `run_test.*`, patches, build scripts and the `meta.yaml` should be moved to a data directory.

@kmuehlbauer don't worry about that in this PR. I was most think out loud what needs to be done here to get this recipe in a better shape.

kmuehlbauer:
@ocefpaf `make -s` isn't that silent as assumed. Maybe we need to redirect output into a file.

ocefpaf:
> @ocefpaf `make -s` isn't that silent as assumed. Maybe we need to redirect output into a file.

I tried that without success in the past, but it is worth a shot.

kmuehlbauer:
> I tried that without success in the past, but it is worth a shot.

I remember that the CI's also don't like absolute quietness and will break the build after some time. Is that correct?

ocefpaf:
> I remember that the CI's also don't like absolute quietness and will break the build after some time. Is that correct?

I don't really recall the issue I had (too long ago), but you are correct and that was probably it. We need to think about this a little further and reduce the noise without suppressing it entirely.

kmuehlbauer:
OK, CircleCI breaks now, because it doesn't find the files:

```yaml
test:
    files:
        - os1_hw.py
        - sites.dbf
        - sites.sbn
        - sites.sbx
        - sites.shp
        - sites.shx
```

Do I need to add the folder?

kmuehlbauer:
> We need to think about this a little further and reduce the noise without suppressing it entirely.

What do you think about something like that: https://github.com/tlatsas/bash-spinner

ocefpaf:
> Do I need to add the folder?

Don't do that now. I need to think about it. Probably change the build test script to find them at the new places. Let's do that in another PR.

@kmuehlbauer I just created the `gdal 1.x` series [branch](https://github.com/conda-forge/gdal-feedstock/tree/gdal_1.x/recipe). Please send this PR to master so the main branch is the latest `gdal`.

kmuehlbauer:
here we go #49

ocefpaf:
@kmuehlbauer let's close this one.

PR #50, opened 2016-05-02
Update gdal to version 2.1.0

kmuehlbauer:


kmuehlbauer:
@ocefpaf Better now, isn't it?

ocefpaf:
> @ocefpaf Better now, isn't it?

Nice and clean :smile:

gillins:
Sorry for delay. I think the following patch should address the Windows build failure.
Unix seems ok since 12 bit isn't the default (have to pass --with-jpeg12 to configure to ask for it). 
Seems by linking to external jpeg has broken 12bit JPEGS. I'm not sure if this is a big issue or not.

```
--- nmake.opt.orig	2016-05-03 08:03:33.803331828 +1000
+++ nmake.opt	2016-05-03 08:03:54.627331424 +1000
@@ -247,7 +247,7 @@
 JPEG_SUPPORTED = 1
 
 # This will enable 12bit libjpeg - use only with internal jpeg builds.
-JPEG12_SUPPORTED = 1
+#JPEG12_SUPPORTED = 1
 
 #if using an external jpeg library uncomment the following lines
 #JPEG_EXTERNAL_LIB = 1
```

kmuehlbauer:
OK, all three OS run fine until the files-test. Since we moved the test-files into `tests`-folder they won't be found. Fixed that with latest commit. 

@gillins There are some `warning LNK4221: no public symbols found; archive member will be inaccessible` within the windows-build also at the nitf-part. Does this any harm?

gillins:
Those warnings aren't serious I don't think - they have always been there as far as I can remember.

kmuehlbauer:
OK, I misunderstand how the test files are working. @gillins do you have experience how this works? If my last commit doesn't work, I'll reset those testfile-changes completely.

kmuehlbauer:
OK, I removed the test-stuff changes completely. Now everything should run fine. :pray:

recipe/meta.yaml
```diff
@@ -11,6 +11,7 @@ source:
     patches:
         - clang.patch  # [osx]
         - prepend-dll.patch  # [win]
+        - disable_jpeg12.patch  # [win]
```
ocefpaf: @kmuehlbauer please add a simple comment here explaining the patch.

kmuehlbauer: @ocefpaf Shall we wait for the CI's to finish and add that comment just before merging this?

recipe/build.sh
```diff
@@ -26,7 +26,7 @@ CPPFLAGS="-I$PREFIX/include" LDFLAGS="-L$PREFIX/lib" \
             --with-python \
             $PGFLAG
 
-make
+make -s
```
ocefpaf: Can you do something along these lines (untested):

```shell
if [[ $(uname) == Darwin ]]; then
 OPTS="-s"  # Silence log to avoid Travis-CI 4 MB limit,
else
 OPTS=""
fi

make
make $OPTS
```

ocefpaf:
@kmuehlbauer I made a few minor comments but this looks good to go!

recipe/meta.yaml
ocefpaf: > @ocefpaf Shall we wait for the CI's to finish and add that comment just before merging this?

Yes. And then you can commit with `[skip ci]` since there are no changes to the build.

kmuehlbauer: > Yes. And then you can commit with [skip ci] since there are no changes to the build.

I didn't get this. Where should I have the [skip ci]? In the commit message?

ocefpaf: >  In the commit message?

Yep. That will prevent `AppVeyor` and `Travis-ci` from starting (it does not work with `CircleCI`, but we can merge it without waiting for it anyways :smile:).  If you ever need to stop only `AppVeyor` there is also `[skip appveyor]`.

kmuehlbauer:
@ocefpaf, @gillins, We have green lights, thanks!

recipe/meta.yaml
```diff
@@ -11,6 +11,8 @@ source:
     patches:
         - clang.patch  # [osx]
         - prepend-dll.patch  # [win]
+        # fixes compilation problems with jpeg
+        # https://lists.osgeo.org/pipermail/gdal-dev/2016-January/043443.html
```
ocefpaf: :+1:

ocefpaf:
Thanks @kmuehlbauer!

`gdal 2` is now our main `gdal` version :tada:

(Still waiting on a new `fiona` release to unpin our stack though :wink:)

kmuehlbauer:
@ocefpaf Just a few questions: Isn't it possible to keep the builds from a successful PR when merging, to not have to reiterate over that again? Was this discussed before? If so, do you have some pointers?

ocefpaf:
> @ocefpaf Just a few questions: Isn't it possible to keep the builds from a successful PR when merging, to not have to reiterate over that again? Was this discussed before? If so, do you have some pointers?

I thought about that some time ago. We could try to save packages as artifacts and have a merge action that would access and upload those artifact. I have no idea what it would take to implement this nor if it is worth doing it...

kmuehlbauer:
> I have no idea what it would take to implement this nor if it is worth doing it...

It is worth doing IMHO, not only for conda-forge, but also for the CI's. But I'm also short of ideas on how to do this due to my lack of knowledge on the conda-forge magic.

ocefpaf:
> It is worth doing IMHO

You are correct. I just meant that it is not worth of our time right now as there are more pressing issues to resolve.

kmuehlbauer:
> I just meant that it is not worth of our time right now as there are more pressing issues to resolve.

Absolutely. If I open an issue about that, to keep hold of this, what would be the best place? https://github.com/conda-forge/conda-forge.github.io

ocefpaf:
> Absolutely. If I open an issue about that, to keep hold of this, what would be the best place? https://github.com/conda-forge/conda-forge.github.io

Yep. https://github.com/conda-forge/conda-forge.github.io is the right place.

Issue #59, opened 2016-05-16
Python 3.4 32-bit failing on Windows/AppVeyor


jakirkham:
Looks like GDAL builds are failing on Python 3.4 32-bit Windows. This started in the last couple builds. Seems to not be able to find `tiffio.h`. Sample failure in this [log]( https://ci.appveyor.com/project/conda-forge/gdal-feedstock/build/1.0.126/job/7ejmd7w95f0ys6qr#L1917 ).

gillins:
I think TIFF_INC=%LIBRARY_INC% -> TIFF_INC=**-I**%LIBRARY_INC% in bld.bat might work. Seems to be missing a -I in front of the include path in the log. Don't know why if works for other combinations.
I'll submit a PR shortly.

gillins:
See https://github.com/conda-forge/gdal-feedstock/pull/60.

I think something changed in the GDAL build between 2.0.2 and 2.1.0, seems the paths are specified differently. I don't know why the other configurations passed.

PR #60, opened 2016-05-16
attempt to fix tiff compilation issue on Windows

gillins:
See error here:
https://github.com/conda-forge/gdal-feedstock/issues/59

Seems that I never had the include set properly for tiff. This now matches the other drivers.


pelson:
Once `Environment: TARGET_ARCH=x86, CONDA_NPY=110, CONDA_PY=34` has passed, I'm 👍 on merging this.

PR #63, opened 2016-05-24
Make gdal activate script also work in other shells.

cpaulik:
See the related issue and PR at https://github.com/conda/conda-recipes/issues/677 and https://github.com/conda/conda-recipes/pull/679

I've tested this locally in bash and zsh.

### DIFF
--- a/recipe/scripts/activate.sh
+++ b/recipe/scripts/activate.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 if [[ -z "$GDAL_DATA" ]]; then
-  DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
-  export GDAL_DATA="${DIR}/../../../share/gdal"
+  export GDAL_DATA=$(gdal-config --datadir)
   export _CONDA_SET_GDAL_DATA=1
 fi
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -17,7 +17,7 @@ source:
         - install_scripts.patch
 
 build:
-    number: 2
+    number: 3
     features:
         - vc9  # [win and py27]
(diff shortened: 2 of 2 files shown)


recipe/scripts/activate.sh
```diff
@@ -1,6 +1,5 @@
 #!/bin/bash
 if [[ -z "$GDAL_DATA" ]]; then
-  DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
-  export GDAL_DATA="${DIR}/../../../share/gdal"
+  export GDAL_DATA=$(gdal-config --datadir)
```
ocefpaf: :+1:

ocefpaf:
Nice addition @cpaulik! Please you bump the build number so this change gets into the release.

ocefpaf:
I cancelled AppvVeyor b/c `gdal` is a heavy build and we don't really need to build it for this change. BTW, once you bump the build number, you can add `[ci skip]` in the commit message so we can merge it ASAP.

gillins:
Should we try to support tcsh also? Just an idle thought...

cpaulik:
I've updated the build number.

ocefpaf:
> Should we try to support tcsh also? Just an idle thought...

Do we have too? With the propper shebang that script will execute correctly in any shell, right?

gillins:
Sourcing doesn't seem to respect the shebang since it must be executed by the same shell. However, I just tested this and `activate` itself doesn't work with tcsh so just ignore my idle thought....

ocefpaf:
Got it. I will investigate if there are any clean solutions to be more inclusive with other shells.

ocefpaf:
@gillins if you agree with the changes here feel free to merge. I will back port it later to `1.11.4`

Issue #66, opened 2016-06-03
Add support for spatialite/sqlite


jgomezdans:
Spatialite (and sqlite) would be very nice to have. There's a conda-foge spatialite package, so this should be as simple as adding

    --with-sqlite3=$PREFIX \
    --with-spatialite=$PREFIX \

to `build.sh`, and to add dependencies to `sqlite` and `libspatialite` in `meta.yaml` (both for compile and run times). This seems to be working on Linux, and I guess it would also work for OSX, and there would be an equivalently simple Windows incantation. I can produce a PR, but this would probably conflict with Windows builds, and the PR is pretty much stated above anyway ;-)

ocefpaf:
@jgomezdans we have [spatialite](https://github.com/conda-forge/gdal-feedstock/blob/gdal_1.x/recipe/build.sh#L32) support in the `gdal 1.11.x` branch and I recall testing sqlite3 and failing (don't remember the failure so I should try again.)

I will put a PR to add both to `gdal 2` as well.

jgomezdans:
So I tried those two changes on gdal2 and "it works for me on Linux". Don't know about OSX or Windows though...

ocefpaf:
Want to send a PR :grin: (Maybe also add yourself as a maintainer :wink:)

jgomezdans:
Forgot to close the issue. It's been solved by merging #68 .

PR #67, opened 2016-06-03
Add support for Spatialite

jgomezdans:
The current changes add support for Spatilate (see #66). Tested
on Linux, and compilation and everything else works. Currently,
the Windows makefile hasn't been changed. The changes are quite
similar to the 1.x branch changes that already support spatialite.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -43,7 +43,7 @@ requirements:
         - postgresql  # [not win]
         - openjpeg
         - curl
-        - sqlite
+        - sqlite  # [not win]
         - libspatialite  # [not win]
 
     run:
@@ -64,7 +64,7 @@ requirements:
         - postgresql  # [not win]
         - openjpeg
         - curl
-        - sqlite 
+        - sqlite  # [not win]
         - libspatialite  # [not win]
 
 test:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -17,7 +17,7 @@ source:
         - install_scripts.patch
 
 build:
-    number: 3
+    number: 4
     features:
         - vc9  # [win and py27]
         - vc10  # [win and py34]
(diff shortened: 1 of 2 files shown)


recipe/meta.yaml
```diff
@@ -58,9 +61,11 @@ requirements:
         - libpng 1.6*
         - zlib 1.2*  # [not win]
         - jpeg 9*
-        - postgresql  # [linux]
+        - postgresql  # [not win]
         - openjpeg
         - curl
+        - sqlite
```
ocefpaf: We can try to modify the `bld.bat` to use this on Windows too, but for now I would like to add the `#  [not win]` here too to avoid giving people the impression that we are using `sqlite` on Windows.

ocefpaf:
I cancelled the AppVeyor build b/c this does not change anything on Windows (and gdal is quite heavy for our CIs).

ocefpaf:
@gillins I am :+1: for this change, but won't merge right now as I want to incorporate @jgomezdans commits here into a new PR I am putting together for `libnetcdf 4.4*` to avoid building `gdal` twice. Are you OK with this plan?

gillins:
Fine by me :smile:

I'll look into Windows support after this is all merged.

ocefpaf:
> I'll look into Windows support after this is all merged.

Thanks. I am closing this in favor of #68 btw

gillins:
@ocefpaf I just had a quick look into getting a Windows build going and found that although `sqlite` is available under Windows, `libspatialite` is not. It doesn't look too hard to do (there is a handy `makefile.vc` for building with Visual Studio) but it depends on `iconv`. `iconv` doesn't currently support building with Visual Studio. From to their `README.woe32`:

```
Building requires the mingw or cygwin development environment (includes gcc). 
MS Visual C/C++ with "nmake" is no longer supported.
```

I can probably work around this, but it might take some work. Thought I'd just let you know.

ocefpaf:
Thanks!

PR #68, opened 2016-06-03
Pin libnetcdf

ocefpaf:
@jgomezdans can you review this to see if everything from your PR is here?

@gillins can you review and merge after @jgomezdans gives the :+1: ?

---
Edit: This PR supersedes #67

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -20,6 +20,9 @@ CPPFLAGS="-I$PREFIX/include" LDFLAGS="-L$PREFIX/lib" \
             --with-jpeg=$PREFIX \
             --with-libtiff=$PREFIX \
             --with-png=$PREFIX \
+            --with-sqlite3=$PREFIX \
+            --with-spatialite=$PREFIX \
+            --with-curl \
             --with-libz=$PREFIX \
             --disable-rpath \
             --without-pam \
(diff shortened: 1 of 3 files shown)


recipe/build.sh
```diff
@@ -20,6 +20,9 @@ CPPFLAGS="-I$PREFIX/include" LDFLAGS="-L$PREFIX/lib" \
             --with-jpeg=$PREFIX \
             --with-libtiff=$PREFIX \
             --with-png=$PREFIX \
+            --with-sqlite3=$PREFIX \
+            --with-spatialite=$PREFIX \
+            --with-curl \
```
ocefpaf: @jgomezdans just curious: why curl here?

jgomezdans: I just had it on a previous version of my build file, so left it over. In all my build, curl is being identified automatically, so this isn't probably needed.

jgomezdans:
Looks good to me, but feel free to get rid of the `--with-curl` in `build.sh` if you think it will cause problems. Otherwise, go ahead and commit ;-)

ocefpaf:
Better explicit than implicit. Let's leave it there. I'll merge once I get
back home if someone does not beat me to it. Thanks again for the
contribution José!
Em 04/06/2016 09:08, "José Gómez-Dans" <notifications@github.com> escreveu:

PR #79, opened 2016-06-26
use toolchain

gillins:
This PR updates the GDAL build to use the toolchain which should fix linking errors on OSX.

ping @jakirkham

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -9,7 +9,6 @@
 source activate "${CONDA_DEFAULT_ENV}"
 
 if [ $(uname) == Darwin ]; then
-    export LDFLAGS="-headerpad_max_install_names"
     OPTS="--enable-rpath"
     export CXX="${CXX} -stdlib=libc++"
 else
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -1,8 +1,17 @@
 #!/bin/bash
 
+# FIXME: This is a hack to make sure the environment is activated.
+# The reason this is required is due to the conda-build issue
+# mentioned below.
+#
+# https://github.com/conda/conda-build/issues/910
+#
+source activate "${CONDA_DEFAULT_ENV}"
+
 if [ $(uname) == Darwin ]; then
     export LDFLAGS="-headerpad_max_install_names"
     OPTS="--enable-rpath"
+    export CXX="${CXX} -stdlib=libc++"
 else
     OPTS="--disable-rpath"
 fi
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -43,7 +43,8 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
             --with-sqlite3=$PREFIX \
             --with-curl \
             --with-python \
-            $OPTS
+            --without-libtool \
+            $OPTS $LIBT
 
 
 if [[ $(uname) == Darwin ]]; then
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -18,7 +18,7 @@ source:
         - install_scripts.patch
 
 build:
-    number: 5
+    number: 6
     features:
         - vc9  # [win and py27]
         - vc10  # [win and py34]
@@ -28,6 +28,7 @@ requirements:
     build:
         - python
         - setuptools
+        - toolchain
         - cmake  # [win]
         - numpy x.x
         - hdf4


gillins:
@jakirkham I'm getting the following error on Travis and locally:
```
libtool: compile: cannot determine name of library object from `build/lib.macosx-10.6-x86_64-3.5/osgeo/_gdal.cpython-35m-darwin.so'
```
Any ideas? Seems to be related to `toolchain` build.

recipe/build.sh
```diff
@@ -1,8 +1,17 @@
...
+# FIXME: This is a hack to make sure the environment is activated.
+# The reason this is required is due to the conda-build issue
+# mentioned below.
+#
+# https://github.com/conda/conda-build/issues/910
+#
+source activate "${CONDA_DEFAULT_ENV}"
+
 if [ $(uname) == Darwin ]; then
     export LDFLAGS="-headerpad_max_install_names"
```
jakirkham: So, this is overriding the `LDFLAGS` set by the `toolchain`. Also, we add this to `LDFLAGS` anyways. Maybe we can try dropping it.

gillins:
Good spotting, but I'm still getting the same problem....

gillins:
Below is the error on Linux (from my machine - seems to get lost in CI). Seems to be confusing `bash` and `gcc`. Weird.
```
/bin/bash -pthread -shared -L/home/sam/miniconda3/envs/_build/lib -Wl,-rpath=/home/sam/miniconda3/envs/_build/lib,--no-as-needed -L/home/sam/miniconda3/envs/_build/lib -L/home/sam/miniconda3/envs/_build/lib -m64 -Wall -Wdeclaration-after-statement -Wextra -Winit-self -Wunused-parameter -Wmissing-prototypes -Wmissing-declarations -Wformat -Werror=format-security -Wno-format-nonliteral -Wlogical-op -Wshadow -Werror=vla -Wdeclaration-after-statement -DOGR_ENABLED -I/home/sam/miniconda3/envs/_build/include -I/home/sam/miniconda3/conda-bld/work/gdal-2.1.0/port -I/home/sam/miniconda3/envs/_build/include -I/home/sam/miniconda3/envs/_build/include -I/home/sam/miniconda3/envs/_build -I/home/sam/miniconda3/envs/_build/include -I/home/sam/miniconda3/envs/_build/include -I/home/sam/miniconda3/envs/_build -I/home/sam/miniconda3/envs/_build/include -I/home/sam/miniconda3/envs/_build -I/home/sam/miniconda3/envs/_build/include -DGDAL_COMPILATION build/temp.linux-x86_64-3.4/extensions/gdal_wrap.o -L../../.libs -L../../ -L/home/sam/miniconda3/envs/_build/lib -L/home/sam/miniconda3/conda-bld/work/gdal-2.1.0/lib -lpython3.4m -lgdal -o build/lib.linux-x86_64-3.4/osgeo/_gdal.cpython-34m.so
/bin/bash: -d: invalid option
error: command '/bin/bash' failed with exit status 1
make[2]: *** [build] Error 1
make[2]: Leaving directory `/home/sam/miniconda3/conda-bld/work/gdal-2.1.0/swig/python'
make[1]: *** [build] Error 2
make[1]: Leaving directory `/home/sam/miniconda3/conda-bld/work/gdal-2.1.0/swig'
```

kmuehlbauer:
@gillins No, not confusing `bash` and `gcc`. 

`libtool` is called this way. Between `/bin/bash` and `-pthread` the path to `libtool` is missing. This is really weird. The first call of `libtool` in `SWIG/python` are working and `gdal_wrap.cpp` is compiled correctly. But the linking goes wrong because `libtool` isn't picked up by the env-variable any more, so it seems at least. Maybe there is something happening to the env-var for `libtool`. I've no idea.

If you run `configure` with `--without-libtool` it compiles well locally on linux.

kmuehlbauer:
https://github.com/LLNL/spack/issues/679 shows the same problem. It might be related to wrongly set environment variables. As this is only with the SWIG bindings, that could be the root cause, that somehow the correct set `libtool` env-var is cleared or overwritten via the python `setup.py`. 

ping @jjhelmus Have you ever seen this libtools-issue, or similar?

kmuehlbauer:
@gillins Seems that CircleCI was skipped too. But the travis runs are OK now, so running without `libtool` works, at least on osx.

kmuehlbauer:
Another thought: Do we need `libtool` as build dependency, if we use it? Or is it already built-in in the docker image.

gillins:
Thanks for the tip @kmuehlbauer but adding `libtool` as a dependency didn't help locally. Seems `--without-libtool` seems the only way to get it to work for now...

ocefpaf:
@gillins and @kmuehlbauer this seems good to go. Thanks! Merging this...

jakirkham:
> Another thought: Do we need libtool as build dependency, if we use it? Or is it already built-in in the docker image.

It's not provided in the docker image. The thought was that if you needed `libtool` it should be a build requirement as we provide it.

On Mac, it could be optional as there is a system copy. Though we have tended towards mimicking what we do on Linux (i.e. adding it as a dependency).

jakirkham:
Would be good to understand what is going on with `libtool`. Is it worth opening an issue?

kmuehlbauer:
@jakirkham We should definitely sort things out. https://github.com/conda-forge/gdal-feedstock/issues/81

jakirkham:
Thanks @kmuehlbauer. I'll keep an eye on that.

PR #82, opened 2016-06-28
remove symlinks

gillins:
I forgot to unpin `conda-build` when using `toolchain` which means symlinks get left in. This should fix it.

cc @jakirkham

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
       export PATH=/Users/travis/miniconda3/bin:$PATH
 
       conda update --yes conda
-      conda install --yes conda-build=1.20.0 jinja2 anaconda-client
+      conda install --yes conda-build jinja2 anaconda-client
       conda config --add channels conda-forge
       conda config --set show_channel_urls true
       
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -44,7 +44,7 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
             --with-curl \
             --with-python \
             --without-libtool \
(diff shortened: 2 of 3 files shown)


jakirkham:
So, this is pretty weird. When I grabbed `osx-64/gdal-2.1.0-np111py35_6.tar.bz2`, it looked fine. Now there are 6 packages. Thus maybe some of them are broken, but it is strange that this one wasn't. 😕

jakirkham:
Could you please let me know which ones are broken, @gillins, and I will mark them as broken?

gillins:
You are right - the latest one (with `toolchain` is fine). Very strange.

Maybe I should remove the build number bump?

jakirkham:
Were all of the OS X build number 6 packages fine?

gillins:
Yes

jakirkham:
😳 Weird. It would be nice if we could narrow down the cause for this change. Maybe it will be easier with `libtiff`? There was a recent `conda` release. So this behavior may have changed with it.

gillins:
Should I close this, and leave it as it was? I'm guessing the next change to `conda-smithy` will update anything that needs to be changed?

jakirkham:
Not sure. In one sense, it isn't a problem here. In the other, we don't really know why it is not a problem or whether it will return. I'd lean towards unpinning because we know that works and why. Though I guess the build number change is optional.

gillins:
Turned out that this package wasn't broken like `hdf5` etc so a rebuild and new version isn't required, but good to have it unpinned in case there is a problem in the future.

Issue #76, opened 2016-06-15
Extra drivers test


ocefpaf:
@gillins in https://github.com/conda-forge/gdal-feedstock/pull/75#issuecomment-225743328

> We need to add these other drivers to `run_test.py`. I was hoping to get `libspatialite` etc going under Windows this week, but I'm heading off on leave until Monday in a few hours, so hopefully next week.

Let's add the exta drivers tests once we have them all working on Windows to avoid workarounds in `run_test.py`.

ocefpaf:
This was addressed in #84 and #90

PR #84, opened 2016-06-30
WIP: Add missing Windows packages and tests

### LINKED ISSUE #74: Add support for expat
As pointed out by @jakirkham, `gdal` is linked against the system `libexpat` on OSX which should be corrected. As we have a conda-forge `libexpat` we should link against that on all platforms which will give us a few more formats.

gillins:
This closes #74 and #76. 

There aren't options to compile with external `zlib`, `giflib` or `json-c` on Windows. I've also skipped `postgresql` since we don't yet have Windows libs for it.

I'm not sure why this package depends on `krb5` - there doesn't seem to be an option to use it. Am I missing something?

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -64,7 +64,15 @@ set BLD_OPTS=%WIN64% ^
     OPENJPEG_LIB=%LIBRARY_LIB%\openjp2.lib ^
     OPENJPEG_VERSION=20100 ^
     CURL_INC="-I%LIBRARY_INC%" ^
-    CURL_LIB="%LIBRARY_LIB%\libcurl.lib wsock32.lib wldap32.lib winmm.lib"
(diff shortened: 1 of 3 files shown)


ocefpaf:
> I'm not sure why this package depends on `krb5` - there doesn't seem to be an option to use it. Am I missing something?

Some time ago another dependency that was build with `krb5` was removed from `defaults` and we added it here as a workaround (could not find the thread on conda's mailing list).  But I think it is safe to remove it now. Want to do it in this PR?

gillins:
We should probably wait until https://github.com/conda-forge/staged-recipes/pull/849 is merged before merging this...

gillins:
Will wait for https://github.com/conda-forge/staged-recipes/pull/937 also....

ocefpaf:
Done in #90

Issue #81, opened 2016-06-27
build problem with libtool


kmuehlbauer:
As experienced in #79, there is currently a problem building gdal python bindings using libtool.

I contacted the gdal-dev list for help in this matter: https://lists.osgeo.org/pipermail/gdal-dev/2016-June/044658.html.

gillins:
I'm now getting a failure on OSX when trying to import `gdal`. Seems the patch to `libgdal.dylib` has been hardcoded:
```
otool -L /Users/sam/miniconda3/envs/myenv/lib/python3.5/site-packages/osgeo/_gdal.cpython-35m-darwin.so
/Users/sam/miniconda3/envs/myenv/lib/python3.5/site-packages/osgeo/_gdal.cpython-35m-darwin.so:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.0.0)
	/Users/travis/miniconda3/conda-bld/work/gdal-2.1.0/libgdal.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/./libstdc++.6.dylib (compatibility version 7.0.0, current version 7.19.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1197.1.1)
```

jakirkham:
...and it seems to still be linked to `libstdc++`. 😒

gillins:
Yep sure is... Not sure if that is related to `libtool`. 

I have been thinking we need a better test for library linking. Would it be possible to add tests for `libstdc++` and libraries under `/Users` etc in an automated way to the build system?

jakirkham:
Certainly worth thinking about.

FWIW, PR ( https://github.com/conda-forge/toolchain-feedstock/pull/9 ) got accepted so the `toolchain` now used 10.9. Maybe this problem already goes away because of that. 🍀

gillins:
That would be good if it does. I'm just testing on my Mac and will report back...

gillins:
With the latest `toolchain`, `libstdc++` goes away, but the absolute path to the build location of `libgdal.dylib` still gets inserted...

jakirkham:
Maybe it is worth trying `libtool` again?

jakirkham:
...or maybe one could try `export LDFLAGS="${LDFLAGS} -Wl,-rpath,$PREFIX/lib"`.

jakirkham:
(after activating the `toolchain` of course)

[...]

PR #85, opened 2016-06-30
WIP: FIX: give CC and CXX as configure parameters

kmuehlbauer:
This should fix the libtool-problem experienced in #79. Discussion went ahead in #81. 

Update:

This fixes some issue with configure and libtool. Obviously CC and CXX exported in the environment break the build. Therefore this PR copies CC and CXX unsets CC and CXX and feeds the copy to the configure script.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -21,7 +21,9 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
 # `--without-pam` was removed.
 # See https://github.com/conda-forge/gdal-feedstock/pull/47 for the discussion.
 
-./configure --prefix=$PREFIX \
+./configure CC=$CC \
+            CXX=$CXX \
+            --prefix=$PREFIX \
             --with-hdf4=$PREFIX \
             --with-hdf5=$PREFIX \
             --with-xerces=$PREFIX \
@@ -43,7 +45,6 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
             --with-sqlite3=$PREFIX \
             --with-curl \
             --with-python \
-            --without-libtool \
             $OPTS
 
 
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -18,11 +18,17 @@ fi
 export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
 
+# make copy and unset CC and CXX
+# see https://github.com/conda-forge/gdal-feedstock/issues/81
+export CF_CC=$CC
+export CF_CXX=$CXX
+unset CC CXX
+
 # `--without-pam` was removed.
 # See https://github.com/conda-forge/gdal-feedstock/pull/47 for the discussion.
 
-./configure CC=$CC \
-            CXX=$CXX \
+./configure CC=$CF_CC \
+            CXX=$CF_CXX \
             --prefix=$PREFIX \
             --with-hdf4=$PREFIX \
             --with-hdf5=$PREFIX \


kmuehlbauer:
OK, this didn't work on CircleCI as it worked locally.

gillins:
@kmuehlbauer See https://help.ubuntu.com/community/EnvironmentVariables the section on "Bash's quick assignment and inheritance trick". Basically you are still passing those values via environment variables - I can't see this making any difference.

kmuehlbauer:
@gillins You are right. I didn't correctly set the env vars. I had removed the `toolchain` and set those vars locally. Made that mistake while cleaning up.

I just ran this configuration locally without trouble. Let's have a look how this behaves in the CI's (the first CircleCi python27 went through :tada:). 

Anyway, having `CC` and `CXX` in the environment with this particular `configure`/`libtool` setup, is breaking things. I know from reading `libtool` and `autoconf` online resources, that there might be issues with environment variables, but I didn't find anything which directly relates to our issue. So making a copy and unsetting CC and CXX .

Pinging @ocefpaf, @jakirkham, @msarahan, @pelson because this might hit other packages too. Any ideas, what is actually going on here?

ocefpaf:
@kmuehlbauer thanks for digging into this. Maybe we could accomplish the same by limiting the `toolchain` to `OS X` with a pre-selector `# [osx]`? That would be cleaner and it is virtually the same things, no?

PS: I am experimenting issue linking with `kealib` using the `toolchain` too and I temporarily removed the latest build number to investigate that. I plan to re-add it if there is no problem, but bare in mind that the latest build here will be using a `kealib` that does not use the `toolchain`!

kmuehlbauer:
> Maybe we could accomplish the same by limiting the toolchain to OS X with a pre-selector # [osx]? That would be cleaner and it is virtually the same things, no?

The `toolchain` also sets the CC, CXX, and more FLAGS for linux. So we would get the if-switch into the build.sh again. Don't know, if that's that easy. Unfortunately travis is full and the runs will be canceled anyway, but maybe @jakirkham or @gillins can test this locally.

ocefpaf:
This is what the `toolchain` does on Linux:

```
    # for Linux
    export CC=gcc
    export CXX=g++
    export CFLAGS="${CFLAGS}"
    # Boost wants to enable `float128` support on Linux by default.
    # However, we don't install `libquadmath` so it will fail to find
    # the needed headers and fail to compile things. Adding this flag
    # tells Boost not to support `float128` and avoids this search
    # process. As it has confused a few people. We have added it here.
    # The idea to add this flag was inspired by this Boost ticked.
    #
    # https://svn.boost.org/trac/boost/ticket/9240
    #
    export CXXFLAGS="${CXXFLAGS} -DBOOST_MATH_DISABLE_FLOAT128"
    export LDFLAGS="${LDFLAGS}"
    export LINKFLAGS="${LDFLAGS}"

export CFLAGS="${CFLAGS} -m${ARCH}"
export CXXFLAGS="${CXXFLAGS} -m${ARCH}"
```

So unless you are using boost or need the -m${ARCH} (I believe only geos needs that) we are OK skipping it on Linux. I am not OK with the `toolchain` as is made things harder to debug and evolve. We need to unset things, test, and then update the `toolchain`, and hope that does not break anything else. That is a bad workflow IMO.

kmuehlbauer:
@ocefpaf Yes, gdal also compiles without the toolchain on linux. *Hard to debug*, nice wording for what I'm through :grimacing: 

I can make the skip tomorrow, or you can try before, if you have time.

kmuehlbauer:
FYI: I cancel travis and appveyor.

ocefpaf:
>  *Hard to debug*

In italics that looks like a nice title for movie :wink: 

> I can make the skip tomorrow, or you can try before, if you have time.

I might try it later today. Just want https://travis-ci.org/conda-forge/kealib-feedstock/builds/140334714 to finish to avoid adding more variables to this problem.

gillins:
OK that does fix it! Well done @kmuehlbauer. Sorry for doubting. Using a `config.site` could be another way around this. 

However the test of the Python bindings fail:
```
ImportError: dlopen(/Users/sam/miniconda3/envs/_test/lib/python3.5/site-packages/osgeo/_gdal.cpython-35m-darwin.so, 2): Library not loaded: /Users/sam/miniconda3/conda-bld/work/gdal-2.1.0/libgdal.dylib
  Referenced from: /Users/sam/miniconda3/envs/_test/lib/python3.5/site-packages/osgeo/_gdal.cpython-35m-darwin.so
  Reason: image not found
TESTS FAILED: gdal-2.1.0-np111py35_6
```
See below (note how the build path is still hardcoded):
```
$ otool -L /Users/sam/miniconda3/envs/_test/lib/python3.5/site-packages/osgeo/_gdal.cpython-35m-darwin.so
/Users/sam/miniconda3/envs/_test/lib/python3.5/site-packages/osgeo/_gdal.cpython-35m-darwin.so:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	/Users/sam/miniconda3/conda-bld/work/gdal-2.1.0/libgdal.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```
Seems to be some weird stuff going on on OSX...

kmuehlbauer:
@gillins The Travis runs indicate that gdal is built correctly for the whole matrix (excluding py27 np111, which timed out). 

So unsetting CC and CXX also worked on OSX. I can't say anything about the libc++-issue, though.

A decision is needed, how to proceed with `toolchain` and setting environment variables. It is not that useful to first set variables in `toolchain` with the need to unset them in another package. ping @ocefpaf, @jkirkham, @pelson

Also asked again in the gdal-dev list giving steps to reproduce the behaviour (https://lists.osgeo.org/pipermail/gdal-dev/2016-July/044678.html).

kmuehlbauer:
Restarted the timed out build, all green on osx now.

ocefpaf:
@kmuehlbauer for what it is worth RPMs have the concept of `optflags` macros to inject flags.
However, different from the `toolchain` the `optflags` are used to:

> While the optflags entry doesn't play a part in determining the build or install platform, it does play a role in multi-platform package building. 

There is some wisdom there. Anything that plays a plays a part in the `build` or `install platform` may force something that is ideal in some cases and not ideal in others. I don't think we have a silver bullet that is worth adding in a `toolchain`. In the long run we may be creating a larger problem by adopting it everywhere.

source: http://www.rpm.org/max-rpm/s1-rpm-multi-optflags.html

kmuehlbauer:
If the use of `toolchain` in the momentary version creates this much problems, then it should be reverted, abandoned (you name it). At least until everything is finally sorted out.

So for `gdal` I propose dropping of `toolchain` here and in all dependencies.

ocefpaf:
> So for `gdal` I propose dropping of `toolchain` here and in all dependencies.

:+1: 

How about you @gillins?

gillins:
I guess so,  but I think we need to resolve the original problem on osx (that is being linked to both libc++ and libstdc++ at the same time). The toolchain did at least begin to resolve that one. 

I don't really understand all the issues, especially on osx so whatever you decide is cool with me.

ocefpaf:
> I don't really understand all the issues, especially on osx so whatever you decide is cool with me.

You do understand it more than me :smile: 

Are you still up? Wanna chat with @pelson now about the issue on google hangout?

ocefpaf:
> I guess so, but I think we need to resolve the original problem on osx (that is being linked to both libc++ and libstdc++ at the same time). The toolchain did at least begin to resolve that one.

I thought that only happened with packages that pulled `openblas`. Or you did see this in `kealib` and `hdf5` too?

gillins:
Sorry bed time for me. Yeah I saw this problem with any c++ library. Just became a problem when openblas came along as suddenly there was a new libstdc++ that some packages were linked against but others were linked against the system one. Turned out they should have linked with libc++ instead. Very weird. I can write up more detail in the morning if you need.

ocefpaf:
> Sorry bed time for me. Yeah I saw this problem with any c++ library. Just became a problem when openblas came along as suddenly there was a new libstdc++ that some packages were linked against but others were linked against the system one. Turned out they should have linked with libc++ instead. Very weird. I can write up more detail in the morning if you need. 

Thanks! That is what I thought.  (Good night and enjoy your weekend!)

jakirkham:
So, I have tried to write up the `libstdc++` saga (yeah sorry it's long) in this issue ( https://github.com/conda-forge/conda-forge.github.io/issues/183 ).

kmuehlbauer:
@jakirkham, @gillins Could you please check, if the libc++ issue is solved with this changes here. It build correctly in the CI.

pelson:
> @jakirkham, @gillins Could you please check, if the libc++ issue is solved with this changes here. It build correctly in the CI.

Is it likely we need to have solved `openblas` pulling in `libstdc++` do you think?

kmuehlbauer:
> Is it likely we need to have solved openblas pulling in libstdc++ do you think?

I'm sorry, but I didn't get this. Since I've no Mac for testing I can't really improve my knowledge in that.

pelson:
Sorry @kmuehlbauer - my comment lacked context. gdal depends on numpy which depends on openblas which depends on libgcc. This may be one package which is causing pollution, but I'm uncertain of that and was hoping for @jakirkham's confirmation/suggestion.

kmuehlbauer:
> gdal depends on numpy which depends on openblas which depends on libgcc. This may be one package which is causing pollution, but I'm uncertain of that and was hoping for @jakirkham's confirmation/suggestion.

This is what @ocefpaf mentioned right from the beginning.

To the specifics of this PR here. Before we had CC and CXX exported in the environment which broke things. Therefore this PR copies CC and CXX unsets, CC and CXX and feeds the copy to the configure script. This might also have impact on linking. This need to be checked. I think @gillins will do this when he is up again.

jakirkham:
> > @jakirkham, @gillins Could you please check, if the libc++ issue is solved with this changes here. It build correctly in the CI.

FWICT looking at one of the builds it looks ok. We should check the rest.

Basically `conda-build` helps us here as it fixes the linkages after the build. So, if `libstdc++` is linked to, you should see it when `install_name_tool` is running. In the build I looked at, it did not show up. 

> Is it likely we need to have solved `openblas` pulling in `libstdc++` do you think?

So there were two issues that came up in issue ( https://github.com/conda-forge/numpy-feedstock/issues/15 ). Here they are.

1. Linux builds are having mixed linkages between the system `libstdc++` and the conda `libgcc` one.
2. OS X builds are using `libstdc++` in some places (sometimes from the system and sometimes not).

There may be more to 1. I have never been totally clear. In any event, FWICT that appears solved by making `openblas` pull in `libgfortran` instead on Linux.

As far as 2 is concerned, this is a different issue. We should be using `libc++` for the reasons discussed in issue ( https://github.com/conda-forge/conda-forge.github.io/issues/183 ). Even if things pull in `libstdc++` via the `libgcc` package it shouldn't matter as we shouldn't be using `libstdc++` at all. So, shadowing shouldn't be an issue at all.

<br>
PS - Not totally relevant, but FWIW `openblas` does still pull in `libstdc++` as there is no `libgfortran` for OS X from conda yet.

gillins:
On my Mac I'm still getting problems loading the Python GDAL bindings, seems to be linked to the build location of `libgdal.dylib` (same as this: https://github.com/conda-forge/fiona-feedstock/pull/17#issuecomment-229328116). I can only assume that this isn't seen on the CI because `conda build` is run with the `--dirty` flag. Would this be correct?

I have no idea what is causing this...

ocefpaf:
> I have no idea what is causing this...

I wonder if we should try reducing the amount of flags added from the `toolchain` to debug this. Maybe all we need on OS X are:

```
    export CXXFLAGS="${CXXFLAGS} -stdlib=libc++"
    export LDFLAGS="${LDFLAGS} -L$PREFIX/lib -headerpad_max_install_names"
    export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
```

recipe/build.sh
```diff
@@ -18,10 +18,18 @@ fi
 export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
 
+# make copy and unset CC and CXX
+# see https://github.com/conda-forge/gdal-feedstock/issues/81
+export CF_CC=$CC
+export CF_CXX=$CXX
+unset CC CXX
```
jakirkham: Yeah, I had to go through some stupid stuff like this with OpenBLAS. ( https://github.com/conda-forge/staged-recipes/pull/364/commits/fed23798ddc8028a16690df848ab805ed105ad92 ) 😒

pelson:
So, other than bumping the build number, is there a reason *not* to merge this PR? The Windows build appears to have been cancelled after one successful build (perfectly reasonable IMO).

ocefpaf:
> So, other than bumping the build number, is there a reason *not* to merge this PR? The Windows build appears to have been cancelled after one successful build (perfectly reasonable IMO).

Strong reasons? No. I believe #87 is another workaround, similar to this one, that may be more promising in the long term. I lost track of which one is better to merge...

BTW: I would like to avoid the `toolchain` so we could avoid the workaround workaround and just copy the flags that are needed here to make the recipe more explicit, easy to debug, and friendlier to  people building `gdal` outside `conda-forge`.

kmuehlbauer:
This here is similar to #87. Here we copy and unset CC and CXX, which have been introduced by the `toolchain`. In #87 config.site mechanism is used to accomplish that. I'm not in favour of workaround stuff which could easily be avoided.

So I back @ocefpaf suggestion to avoid `toolchain`.

pelson:
OK, thanks @kmuehlbauer. Let's close this, and focus on #87 if that is OK?

kmuehlbauer:
Closing in favour of #87

PR #90, opened 2016-07-12
No toolchain solution

ocefpaf:
Jus testing... Nothing to see here. Move along!

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -1,18 +1,48 @@
 #!/bin/bash
 
-# FIXME: This is a hack to make sure the environment is activated.
-# The reason this is required is due to the conda-build issue
-# mentioned below.
-#
-# https://github.com/conda/conda-build/issues/910
-#
-source activate "${CONDA_DEFAULT_ENV}"
+set -e # Abort on error.
+
+export PING_SLEEP=30s
+export WORKDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+export BUILD_OUTPUT=$WORKDIR/build.out
+
+touch $BUILD_OUTPUT
+
+dump_output() {
+   echo Tailing the last 500 lines of output:
+   tail -500 $BUILD_OUTPUT
+}
+error_handler() {
+  echo ERROR: An error was encountered with the build.
+  dump_output
+  exit 1
+}
+
+# If an error occurs, run our error handler to output a tail of the build.
+trap 'error_handler' ERR
+
+# Set up a repeating loop to send some output to Travis.
+bash -c "while true; do echo \$(date) - building ...; sleep $PING_SLEEP; done" &
+PING_LOOP_PID=$!
+
+## START BUILD
+# We could pin to conda-forge's `libxml2`, but this has the same effect and
+# allow us to use either conda-forge of defaults version.
+rm -rf $PREFIX/lib/*.la
 
 if [ $(uname) == Darwin ]; then
+    export LDFLAGS="-headerpad_max_install_names"
     OPTS="--enable-rpath"
-    export CXX="${CXX} -stdlib=libc++"
+    export CXXFLAGS="-stdlib=libc++ $CXXFLAGS"
+    COMP_CC=clang
+    COMP_CXX=clang++
+    export MACOSX_DEPLOYMENT_TARGET="10.9"
+    export CXXFLAGS="${CXXFLAGS} -stdlib=libc++"
+    export LDFLAGS="${LDFLAGS} -headerpad_max_install_names"
 else
     OPTS="--disable-rpath"
+    COMP_CC=gcc
+    COMP_CXX=g++
 fi
 
 export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
@@ -21,7 +51,9 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
 # `--without-pam` was removed.
 # See https://github.com/conda-forge/gdal-feedstock/pull/47 for the discussion.
 
-./configure --prefix=$PREFIX \
+./configure CC=$COMP_CC \
+            CXX=$COMP_CXX \
+            --prefix=$PREFIX \
             --with-hdf4=$PREFIX \
             --with-hdf5=$PREFIX \
             --with-xerces=$PREFIX \
@@ -43,18 +75,10 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
             --with-sqlite3=$PREFIX \
             --with-curl \
             --with-python \
-            --without-libtool \
             $OPTS
 
-
-if [[ $(uname) == Darwin ]]; then
-    OPTS="-s"  # Silence log to avoid Travis-CI 4 MB limit.
-else
-    OPTS=""
-fi
-
-make $OPTS
-make install
+make >> $BUILD_OUTPUT 2>&1
+make install >> $BUILD_OUTPUT 2>&1
 
 # Make sure GDAL_DATA and set and still present in the package.
 # https://github.com/conda/conda-recipes/pull/267
@@ -65,3 +89,10 @@ mkdir -p $DEACTIVATE_DIR
 
 cp $RECIPE_DIR/scripts/activate.sh $ACTIVATE_DIR/gdal-activate.sh
 cp $RECIPE_DIR/scripts/deactivate.sh $DEACTIVATE_DIR/gdal-deactivate.sh
+## END BUILD
+
+# The build finished without returning an error so dump a tail of the output.
+dump_output
+
+# Nicely terminate the ping output loop.
+kill $PING_LOOP_PID
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -18,7 +18,7 @@ source:
         - install_scripts.patch
 
 build:
-    number: 6
+    number: 7
     features:
         - vc9  # [win and py27]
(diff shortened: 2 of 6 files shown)


recipe/meta.yaml
```diff
@@ -28,7 +28,7 @@ requirements:
     build:
         - python
         - setuptools
-        - toolchain
+        - mkl
```
ocefpaf: @msarahan is there a way to ensure that `numpy` with `mkl` is installed? This does not seem to work.

ocefpaf:
Getting somewhere... Both deleting the `.la` files or forcing our `libxml2` gets us one step closer and reveals another issue (I believe with `kealib-1.4.6-2 `):

```shell
Undefined symbols for architecture x86_64:
  "kealib::KEAImageIO::isKEAImage(std::string)", referenced from:
      KEADataset::Identify(GDALOpenInfo*) in keadataset.o
  "kealib::KEAImageIO::openKeaH5RW(std::string, int, unsigned long long, unsigned long long, double, unsigned long long, unsigned long long)", referenced from:
      KEADataset::Open(GDALOpenInfo*) in keadataset.o
      KEADataset::CreateCopy(char const*, GDALDataset*, int, char**, int (*)(double, char const*, void*), void*) in keadataset.o
  "kealib::KEAImageIO::createKEAImage(std::string, kealib::KEADataType, unsigned int, unsigned int, unsigned int, std::vector<std::string, std::allocator<std::string> >*, kealib::KEAImageSpatialInfo*, unsigned int, unsigned int, int, unsigned long long, unsigned long long, double, unsigned long long, unsigned long long, unsigned int)", referenced from:
      KEADataset::CreateLL(char const*, int, int, int, GDALDataType, char**) in keadataset.o
  "kealib::KEAImageIO::openKeaH5RDOnly(std::string, int, unsigned long long, unsigned long long, double, unsigned long long, unsigned long long)", referenced from:
      KEADataset::Open(GDALOpenInfo*) in keadataset.o
  "kealib::KEAImageIO::setImageMetaData(std::string, std::string)", referenced from:
      KEACopyMetadata(GDALMajorObject*, kealib::KEAImageIO*, int) in keacopy.o
      KEADataset::SetMetadataItem(char const*, char const*, char const*) in keadataset.o
      KEADataset::SetMetadata(char**, char const*) in keadataset.o
  "kealib::KEAImageIO::setImageBandMetaData(unsigned int, std::string, std::string)", referenced from:
      KEARasterBand::SetMetadataItem(char const*, char const*, char const*) in keaband.o
      KEARasterBand::SetMetadata(char**, char const*) in keaband.o
      KEACopyMetadata(GDALMajorObject*, kealib::KEAImageIO*, int) in keacopy.o
  "kealib::KEAImageIO::setImageBandDescription(unsigned int, std::string)", referenced from:
      KEARasterBand::SetDescription(char const*) in keaband.o
      KEACopyDescription(GDALRasterBand*, kealib::KEAImageIO*, int) in keacopy.o
  "kealib::KEAImageIO::setGCPs(std::vector<kealib::KEAImageGCP*, std::allocator<kealib::KEAImageGCP*> >*, std::string)", referenced from:
      KEACopyGCPs(GDALDataset*, kealib::KEAImageIO*) in keacopy.o
      KEADataset::SetGCPs(int, GDAL_GCP const*, char const*) in keadataset.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[1]: *** [libgdal.la] Error 1
make: *** [check-lib] Error 2
Command failed: /bin/bash -x -e /Users/travis/build/ocefpaf/gdal-feedstock/recipe/build.sh
```

Testing with another `kealib` now see we can get any further...

jakirkham:
So, the error you have shown above is the sort of error one sees when one builds two libraries with two different incompatible C++ libraries and tries to link them together. I don't recall all of the things that were broken between `libc++` and `libstdc++`, but one of them was `std::string`. Notice every missing symbol shown above uses `std::string`.

recipe/build.sh
```diff
@@ -1,16 +1,10 @@
 #!/bin/bash
 
-# FIXME: This is a hack to make sure the environment is activated.
-# The reason this is required is due to the conda-build issue
-# mentioned below.
-#
-# https://github.com/conda/conda-build/issues/910
-#
-source activate "${CONDA_DEFAULT_ENV}"
+# rm -rf $PREFIX/lib/*.la
```
ocefpaf: @gillins and @kmuehlbauer this is a radical solution, but it saves us from the bad libtool archive files in `defaults`. The alternative is to pin `libxml2 to 2.6.3` to ensure that only the conda-forge version will be installed.

recipe/build.sh
```diff
@@ -1,16 +1,10 @@
...
 
 if [ $(uname) == Darwin ]; then
+    export LDFLAGS="-headerpad_max_install_names"
     OPTS="--enable-rpath"
-    export CXX="${CXX} -stdlib=libc++"
```
ocefpaf: @gillins if the PR to create a `libgfortran` a is merged we won't need this nor the `toolchain` b/c there is no other source for `libstdc++` on the system (`gcc` from `homebrew` is already removed at the beginning of he CI run).

Let's wait for that PR, but you can try https://github.com/conda-forge/gdal-feedstock/pull/90/files#r70559805 in PR #87 and, if that works, we can merge it to finally get a `gdal` that works.

jakirkham: There is still the system `libstdc++`. 😒

kmuehlbauer:
@ocefpaf Fingers crossed here :four_leaf_clover:

ocefpaf:
> @ocefpaf Fingers crossed here :four_leaf_clover: 

Getting there... Things are passing, but there are still a few kinks I am working in the recipe.

kmuehlbauer:
I'm wondering why the conda-forge `numpy` (openblas) is pulled in in build, but in test the default channel numpy is pulled in.

ocefpaf:
> I'm wondering why the conda-forge `numpy` (openblas) is pulled in in build, but in test the default channel `numpy` is pulled in.

That was the main source of headache for us b/c the `numpy`+`openblas` brings `libgcc`, but when you don't get it at run time things are broken. In theory that is no longer a problem since we are only using the `libgfortran` on both `numpy` and `scipy`. However, I always force replace `conda-forge`'s `numpy` with `defaults` anyways b/c `conda-forge`'s `openblas` is single threaded and quite slow in comparison.

See https://github.com/conda-forge/openblas-feedstock/pull/9

jakirkham:
> I'm wondering why the conda-forge `numpy` (openblas) is pulled in in build, but in test the default channel numpy is pulled in.

This is a very old bug in `conda-build` and I have been making noise about it for some time. Maybe now it can get the attention it deserves. It would require some similar sort of workaround as we see with Windows and `python` to fix it. However, it would be a bit uglier as it is not `python`, but `blas` that we are trying to control.

ocefpaf:
@kmuehlbauer and @gillins if this passes feel free to do the honors and merge it :tada:

kmuehlbauer:
@ocefpaf CircleCi has timed out (120min) in previous builds, I'm afraid it won't build in time :cry:

ocefpaf:
> @ocefpaf CircleCi has timed out (120min) in previous builds, I'm afraid it won't build in time :cry:  

That is not a problem.If some of the matrix items got build we can still merge it and then keep re-starting it to complete the rest.

kmuehlbauer:
> @kmuehlbauer and @gillins if this passes feel free to do the honors and merge it :tada:

Sure! Will do, if this is still open after :zzz: 

And, thanks for creating this solution  +:100: .

ocefpaf:
@kmuehlbauer the problem with `CircleCi` might be my workaround to avoid the `4 mb` issue with `Travis-CI` (which BTW came back :unamused:). If AppVeyor passes let's get this merge so we can fix `OS X` and I will worry about `CircleCI` later.

kmuehlbauer:
@ocefpaf I think your solution to silence the build is good. I proposed similar thing a while back, but it got lost somehow. The problem is that the current build of the whole matrix does not fit into the 120 min limit. Can we build with 2 cores to speed up things?

ocefpaf:
> @ocefpaf I think your solution to silence the build is good. I proposed similar thing a while back, but it got lost somehow. The problem is that the current build of the whole matrix does not fit into the 120 min limit. Can we build with 2 cores to speed up things?

In theory we are already using that b/c someone added a flag somewhere to do that. Sorry for been vague, but people are adding all these layers on top of layers and things are becoming really hard to follow and debug.

I will open another PR to test this soon.

kmuehlbauer:
> In theory we are already using that b/c someone added a flag somewhere to do that. Sorry for been vague, but people are adding all these layers on top of layers and things are becoming really hard to follow and debug.

Yes, I saw this `conda-forge-build-setup`, but is this actually used? Anyway, *keep things simple* seems the right approach.

kmuehlbauer:
> I will open another PR to test this soon.

You might try the bugfix-release `gdal 2.1.1`. Or we get this in, when this is all sorted out.

ocefpaf:
Yes, I saw this `conda-forge-build-setup`, but is this actually used? Anyway, *keep things simple* seems the right approach.

Yep. The KISS approach :stuck_out_tongue_winking_eye: 

> You might try the bugfix-release `gdal 2.1.1`. Or we get this in, when this is all sorted out.

Working on it now. I had a few requests for it here at SciPy. (And there is even a `1.11.5`!)

Issue #91, opened 2016-07-13
Symbol not found: __ZN6kealib10KEAImageIO10isKEAImageESs


BtbN:
This looks like gdal needs a rebuild against the latest version of kealib?

https://travis-ci.org/OGGM/OGGM-Anaconda/jobs/144452309#L1940

ocefpaf:
@BtbN I tested this locally and it seems to be fixed with in the latest build. Let me know if you are still finding problems.

Issue #83, opened 2016-06-29
vsizip behaves strangely: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".


akorosov:
**Problem**
After Installing gdal from conda-forge on fresh linux and attempting to open zip file with Sentinel data I got the following errors:

```
gdalinfo /vsizip//vagrant/shared/test_data/sentinel1a_l1/S1A_EW_GRDM_1SDH_20150702T172954_20150702T173054_006635_008DA5_55D1.zip/S1A_EW_GRDM_1SDH_20150702T172954_20150702T173054_006635_008DA5_55D1.SAFE/measurement/s1a-ew-grd-hv-20150702t172954-20150702t173054-006635-008da5-002.tiff
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
Warning 1: Recode from CP437 to UTF-8 failed with the error: "Invalid argument".
[...]

PR #93, opened 2016-07-18
Bump to 2.1.1

ocefpaf:
Testing latest version.

PS: Maybe we should do `1.11.5` too...

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -77,7 +77,8 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
             --with-python \
             $OPTS
 
-make >> $BUILD_OUTPUT 2>&1
+# CircleCI offers two cores.
+make -j 2 >> $BUILD_OUTPUT 2>&1
 make install >> $BUILD_OUTPUT 2>&1
 
 # Make sure GDAL_DATA and set and still present in the package.
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -5,9 +5,9 @@ package:
     version: {{ version }}
 
 source:
-    fn: gdal-{{ version }}.tar.xz
-    url: http://download.osgeo.org/gdal/{{ version }}/gdal-{{ version }}.tar.xz
-    sha256: 87ce516ce757ad1edf1e21f007fbe232ed2e932af422e9893f40199711c41f92
+    fn: gdal-{{ version }}.tar.gz
+    url: http://download.osgeo.org/gdal/{{ version }}/gdal-{{ version }}.tar.gz
+    sha256: 55fc6ffbe76e9d2e7e6cf637010e5d4bba6a966d065f40194ff798544198236b
     patches:
         - clang.patch  # [osx]
         - prepend-dll.patch  # [win]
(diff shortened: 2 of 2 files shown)


kmuehlbauer:
@gillins Any idea why in the windows python 2.7 builds the xz-archives cannot be extracted?

gillins:
Yikes. I've got a terrible feeling this is related to https://github.com/conda-forge/libxml2-feedstock/pull/15. I thought there was a reason that we weren't to upgrade, but I couldn't remember the details - maybe something to do with that version of `xz` not building with VS2008? @ocefpaf will know I hope...

recipe/build.sh
```diff
@@ -77,7 +77,8 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
             --with-python \
             $OPTS
 
-make >> $BUILD_OUTPUT 2>&1
+# CircleCI offers two cores.
+make -j 2 >> $BUILD_OUTPUT 2>&1
```
kmuehlbauer: Neat, this did the trick on Circle wrt to maximum build time (120m).

gillins:
Seems that `xz` 5.2 is VS 2015 (Python 3.5) only: https://github.com/bminor/xz/blob/master/windows/INSTALL-MSVC.txt, https://github.com/conda-forge/xz-feedstock/blob/master/recipe/meta.yaml

ocefpaf:
> Seems that `xz` 5.2 is VS 2015 (Python 3.5) only: https://github.com/bminor/xz/blob/master/windows/INSTALL-MSVC.txt, https://github.com/conda-forge/xz-feedstock/blob/master/recipe/meta.yaml

Ouch! And I thought I was being smart by using a smaller tarball.

gillins:
Ah right - it seems `libxml2` only depends on `xz` on Unix, not Windows so you are right - not a problem.

ocefpaf:
Wow! `AppVeyor` is finishing before `CircleCI` after the upgrade!

kmuehlbauer:
> Wow! AppVeyor is finishing before CircleCI after the upgrade!

You are referring to AppVeyor's recent move to RackSpace? They did a great job, this is a real improvement.

ocefpaf:
> You are referring to AppVeyor's recent move to RackSpace? They did a great job, this is a real improvement.

Yep. It is still a single, but the infrastructure is much better.

ocefpaf:
We have #91 and #83 opened, so I won't merge this before I can check if I can fix those here to avoid multiple builds in the channel.

kmuehlbauer:
> We have #91 and #83 opened, so I won't merge this before I can check if I can fix those here to avoid multiple builds in the channel.

Sure, #91 might be already fixed due to new kealib version, and you also did a fiona-update recently, @ocefpaf?

Don't know how to tackle #83, though.

ocefpaf:
> Sure, #91 might be already fixed due to new kealib version

I am in transit today, but as soon as I settle in the airport I will confirm that.

> and you also did a fiona-update recently, @ocefpaf?

Yep. Should be working. If not let me know :grimacing: 

> Don't know how to tackle #83, though.

I want to at least add a test to understand what is going on.

pelson:
LGTM 👍

ocefpaf:
I was waiting on more comments in #95, but I guess there is no point. We either have to split like `defaults` or keep it in as 1 package. I will send another PR for the split later next week.

PR #105, opened 2016-07-29
Add a proper test for vsizip

ocefpaf:
Add an actual test that reproduces #83


gillins:
Do we need to do something for Windows also? I'm guessing the same problem exists there also.

ocefpaf:
It didn't on `OS X` (see the first commit), but I should probably create a `run_test.bat` with the tests in `run_test.sh` to be sure too. (I can do that tomorrow.)

ocefpaf:
@gillins I probably messed up the `gdalwarp` command in the `bld.bat`, but the others are working!

PR #108, opened 2016-08-10
Add postgresql support to Windows builds

gillins:


ocefpaf:
Thanks @gillins for the PR!

@msarahan this is possible thanks to your hard work on `postgresql`! Thanks again for this!!

ocefpaf:
The following got stuck on `AppVeyor` at `import: 'osr'`:

```
Environment: TARGET_ARCH=x64, CONDA_NPY=110, CONDA_PY=34
Environment: TARGET_ARCH=x64, CONDA_NPY=110, CONDA_PY=35
Environment: TARGET_ARCH=x64, CONDA_NPY=111, CONDA_PY=35
```

Not sure if that was intermittent and restarting will solved or if it is a deeper problem.

gillins:
I think there is a deeper problem - see https://ci.appveyor.com/project/conda-forge/gdal-feedstock/build/1.0.223/job/q67j48rc60rfq705#L148 and note how there is a mix of vc9 and vc14 packages. They should be all vc14 for this build.
I was actually getting this problem here, but thought there was something wrong with my setup. Is this a known problem with the latest `conda-build`?

ocefpaf:
>I think there is a deeper problem - see https://ci.appveyor.com/project/conda-forge/gdal-feedstock/build/1.0.223/job/q67j48rc60rfq705#L148 and note how there is a mix of vc9 and vc14 packages. They should be all vc14 for this build.

Ouch! @msarahan any advice here?

> I was actually getting this problem here, but thought there was something wrong with my setup. Is this a known problem with the latest `conda-build`?

Don't know... We can try down grading it to confirm. I would try `conda-build=1.21.7` first.

gillins:
Sorry for the delay. Same problem with `conda-build=1.21.7`. Should I be trying anything else?

ocefpaf:
No idea... @patricksnape or @msarahan any clue?

patricksnape:
My gut feeling is something wasn't build properly with respect to the `vc` features? But unfortunately that requirements list is enormous so going through each one to check would suck. I'm not really sure how the other `vc` feature is getting activated but as far as I know there's actually nothing preventing that in the features architecture at the moment - so it's pretty difficult to 'force' it to stop.

gillins:
Seems only a problem when we are using `postgresql` - build is fine without. I see that `postgresql` uses the `vc` metapackage, but I don't think any of the other dependencies do. I wonder if this could be causing the problem? (both `vc` and `python` have `track_features` - is this likely to be a problem?)

patricksnape:
I feel like that should be fine because of the Python version being used? This smells like a bug for dependency resolution with regards to features to me...

msarahan:
As long as features are getting activated the way they should, then vc presence or absence should not matter. It's just a means to an end. Adding it elsewhere might be a good idea, as it will just make things easier to reason about.

patricksnape:
@msarahan Maybe we should try removing them from here to see what happens? My gut feeling is that this is either a bug in dependency resolution (some deps are being satisfied with the wrong `vc` feature before the Python enforces the correct one) or some recipe has incorrect `vc` features that aren't properly pre-processed to ensure they only are activated for certain Python versions?

msarahan:
I would add it here to give the right one.

msarahan:
By it, I mean the vc package.

recipe/meta.yaml
```diff
@@ -18,7 +18,7 @@ source:
         - install_scripts.patch
 
 build:
-    number: 1
+    number: 2
     features:
```
patricksnape: @msarahan As in, instead of features, add the actual `vs_runtime` in the build requirements?

msarahan:
@patricksnape, @gillins did what I meant. 

vc is a package. It is versioned. Using it as a dependency activates the corresponding feature. 

vc9, vc10 and vc14 are features. 

Subtle, but important differences. Clear?

gillins:
That seems to have fixed it on Windows. Seems something that might have to be done in other places too.

I've changed the runtime dependency from `postgresql` to `libpq` so users don't have to have the whole database installed. Let me know if this is not ok.

Seems there is a problem with `conda inspect` on Linux and OSX (see https://travis-ci.org/conda-forge/gdal-feedstock/jobs/153451644#L4557). Is this a known problem?

patricksnape:
@msarahan Ah I didn't realise! Good to know! Glad that fixed Windows :+1: Thanks @msarahan!

ocefpaf:
> Seems there is a problem with conda inspect on Linux and OSX (see https://travis-ci.org/conda-forge/gdal-feedstock/jobs/153451644#L4557). Is this a known problem?

I guess this is due the new changes in `conda-build`. We can comment out that test for now. Unless @msarahan knows how to circumvent that :smile:

ocefpaf:
Thanks @gillins!

@msarahan let us know when we can test `conda inspect` here again. That is very useful in `gdal`'s build!

msarahan:
I can't tell from the log what version of conda-build is in use here, and especially whether it is a conda-forge build or not.  How can I see that information?

ocefpaf:
Probably latest from conda-forge. We avoid the output here to workaround
Travis-CI log limit.

I'll check this locally and get back to you with better info.

PR #121, opened 2016-10-24
Test GPKG

ocefpaf:
Test GPKG driver.

Ping @mwengren who reported this to me. It is failing on his Windows machine.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -88,6 +88,16 @@ install:
     - cmd: rmdir C:\cygwin /s /q
     - appveyor DownloadFile "https://raw.githubusercontent.com/pelson/Obvious-CI/master/bootstrap-obvious-ci-and-miniconda.py"
     - cmd: python bootstrap-obvious-ci-and-miniconda.py %CONDA_INSTALL_LOCN% %TARGET_ARCH% %CONDA_PY:~0,1% --without-obvci
+
+    # Add a hack to switch to `conda` version `4.1.12` before activating.
+    # This is required to handle a long path activation issue.
+    # Please see PR ( https://github.com/conda/conda/pull/3349 ).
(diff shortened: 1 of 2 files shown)


ocefpaf:
The Linux failure is due to the recent changes in `libdap4`. Let's hold this PR at the moment until we sort https://github.com/conda-forge/libdap4-feedstock/pull/10 out.

ocefpaf:
@mwengren we are between a rock and a hard place here. `conda-forge` packages are broken on windows until `conda-forge` moves to `conda-build 2` due to https://github.com/conda-forge/hdf5-feedstock/issues/41

In the `IOOS` channel I added a re-build of almost everything that is necessary but we are limited on Windows because of https://github.com/conda/conda-build/issues/1412#issuecomment-256006786

That issue seems to be escaping us since it was fixed twice now (https://github.com/conda/conda-build/issues/1413 and https://github.com/conda/conda-build/pull/1473) and it is still there :grimacing:

Unfortunately `gdal` from `defaults` is not an option, since the `GPKG` driver you need (tested here and proven to be working) fails there.

I will try to set a Windows machine today and investigate that `conda-build` bug myself as it may be specific to that patch or caused by `conda-build-all`.

cc @msarahan who help a lot with the `conda-build` bug.

mwengren:
@ocefpaf Ok, no worries.  I can wait for a fix for this to arise.  In the meantime I can get by without being able to write to GPKG.  Hope you can work all these issues out because it sounds like there are a lot of downstream affects of the `conda build 2` problem (including the Anaconda/IOOS channel package conflicts I saw at first).  This conda world is new to me so unfortunately the rest of the details are clear as mud to me, but I'm sure it's in good hands!

PR #129, opened 2016-12-12
Regenerate python bindings

v0lat1le:
I seem to have encountered the issue described in https://github.com/swig/swig/issues/567.

in python 3.5
``` python
from osgeo import ogr

def make_geom():
    geom = ogr.Geometry(ogr.wkbPoint)
    geom.AddPoint_2D(0, 0)
    return geom

def gen_list(N):
    for i in range(N):
        geom = make_geom()
        yield i

N = 10
print('gen', list(gen_list(N)))
```
produces `SystemError: <built-in function delete_Geometry> returned a result with an error set`

The issue have been resolved in swig 3.0.8 but the bindings that come with GDAL sources are generated with SWIG 2.0.12.

Regenerating the bindings with the latest swig resolves the issue, but also adds libpcre as a run-time dependency.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -78,6 +78,12 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
             --with-dods-root=$PREFIX \
             $OPTS
 
+# Regenerate python bindings.
+cd swig/python
+make veryclean
+make generate
+cd ../..
+
 # CircleCI offers two cores.
 make -j 2 >> $BUILD_OUTPUT 2>&1
 make install >> $BUILD_OUTPUT 2>&1
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -85,7 +85,7 @@ make generate
 cd ../..
 
 # CircleCI offers two cores.
-make -j 2 >> $BUILD_OUTPUT 2>&1
+make -j $CPU_COUNT >> $BUILD_OUTPUT 2>&1
 make install >> $BUILD_OUTPUT 2>&1
 
 # Make sure GDAL_DATA and set and still present in the package.
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -88,9 +88,11 @@ copy *.lib %LIBRARY_LIB%\ || exit 1
 if errorlevel 1 exit 1
 
 :: Python bindings
-cd swig\python
-%PYTHON% setup.py build
+cd swig
+nmake /f makefile.vc python %BLD_OPTS%
 if errorlevel 1 exit 1
+
+cd python
 
 %PYTHON% setup.py install
 if errorlevel 1 exit 1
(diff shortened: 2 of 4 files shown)


recipe/build.sh
```diff
@@ -78,6 +78,12 @@ export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
...
             $OPTS
 
+# Regenerate python bindings.
+cd swig/python
+make veryclean
+make generate
+cd ../..
+
 # CircleCI offers two cores.
 make -j 2 >> $BUILD_OUTPUT 2>&1
```
ocefpaf: Can you help me with another issue and change the `2` to `$CPU_COUNT` here? (`gdal` takes so long to build that I do not want to send a separate PR to fix that.)

ocefpaf:
You need to bump the build number to get a new build uploaded on the channel.

ocefpaf:
OS X is failing with:

```shell
sed: illegal option -- -
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
/bin/sh: /bin/true: No such file or directory
```

and we don't have the Linux test for now due to a failure in `CircleCI`. I am testing Linux locally with our docker image. Will post the results soon.

gillins:
Is it possible to do the same for Windows?

v0lat1le:
I'm not sure what's up with the `sed` failure... Is it something I broke?

Don't have a local windows setup just yet, so I might have to abuse the CI while I get that going

ocefpaf:
@v0lat1le can you add a test to ensure future updates won't break this again? You example in the first message is OK BTW.

gillins:
> I'm not sure what's up with the sed failure... Is it something I broke?

I think OSX uses BSD `sed` rather than GNU `sed` and they have different options... I'm guessing this part of building `gdal` hasn't been tested on OSX. Is there a way of installing GNU `sed` with Conda? Or patching the makefile to work with BSD `sed`?

ocefpaf:
@v0lat1le AppVeyor is stuck and we are trying to figure out what happened. Meanwhile do you mind adding the test `run_test.py` file?

v0lat1le:
[GDAL updated](https://trac.osgeo.org/gdal/ticket/6749) the pre-generated bindings, so the next version (2.2+) will not have this specific issue.

ocefpaf:
> [GDAL updated](https://trac.osgeo.org/gdal/ticket/6749) the pre-generated bindings, so the next version (2.2+) will not have this specific issue.

Thanks @v0lat1le! Even though that will be in the next release I don't think it hurts to always re-generate them here and test for consistency.

PS: Consider adding yourself as a maintainer to this feedstock :wink:

ocefpaf:
Thanks @v0lat1le!

PR #130, opened 2016-12-12
Swig

ocefpaf:
Just testing...

Ping @gillins

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -18,13 +18,14 @@ source:
     - install_scripts.patch
 
 build:
-  number: 2
+  number: 3
 
 requirements:
   build:
     - python
     - swig
     - setuptools
+    - posix  # [win]
     - cmake  # [win]
     - numpy x.x
     - hdf4
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -23,6 +23,7 @@ build:
 requirements:
   build:
     - python
+    - swig
     - setuptools
     - cmake  # [win]
     - numpy x.x
@@ -75,6 +76,7 @@ requirements:
     - expat
     - sqlite 3.13.*
     - libspatialite
+    - pcre
     - vc   9  # [win and py27]
     - vc  10  # [win and py34]
     - vc  14  # [win and py35]
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+* text=auto
+
+meta.yaml text eol=lf
+build.sh text eol=lf
+bld.bat text eol=crlf
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,13 @@ install:
       conda install --yes --quiet conda-forge-build-setup
       source run_conda_forge_build_setup
 
+      # install conda-build 2.x to build with a long prefix
+      conda install --yes --quiet conda-build=2
+      conda info
+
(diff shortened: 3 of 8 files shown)


recipe/meta.yaml
```diff
@@ -18,12 +18,14 @@ source:
...
 build:
-  number: 2
+  number: 3
 
 requirements:
   build:
     - python
+    - swig
     - setuptools
+    - posix  # [win]
```
ocefpaf: Not sure if we need that to run `make`. @mingwandroid is that correct? (Sorry to keep bothering you with `msys2` questions. I will learn eventually :smile:)

gillins:
`/bin/sh: C:Minicondaconda-bldrecipe_1481580663401_b_envpython.exe: command not found`
doesn't look good (see https://ci.appveyor.com/project/conda-forge/gdal-feedstock/build/1.0.273/job/x53k0t31lywciopt#L4421)

Needs quoting or something here. This could get painful...

gillins:
OK it seems that the `GNUmakefile` uses the $PYTHON environment variable if it exists - then obviously there is a quoting problem with `msys2`. I wonder if you can do something like this to workaround it:
```
cd swig\python
set OLDPYTHON=%PYTHON%
set PYTHON=python
make veryclean
make generate
set PYTHON=%OLDPYTHON%
```
If we can't assume that the `python` on the %PATH% is the one to use then we may have to capture the output of `cygpath %PYTHON%` or something....

mingwandroid:
The problem is not quoting, because no amount of quotes will fix it. What happens is that bash interprets the standard Windows dirsep `\` as the escape character. More than that, the msys2 POSIX subsystem has a different filesystem namespace (it has for example `/etc/fstab`)

Because of this, when calling any POSIX program (make.exe from `m2-make` in this case) from bld.bat you need to convert any env. vars that will be interpreted in the POSIX filesystem namespace to that namespace (in the other direction, MSYS2 does this for you automatically though `MSYS2_ENV_CONV_EXCL` can be used to selectively disable that).

You need to use the following to convert things that contain filepaths:
```
for /F "delims=" %%i in ('cygpath.exe -u %PREFIX%') do set "PREFIX=%%i"
for /F "delims=" %%i in ('cygpath.exe -u -p "%LIB%"') do set "LIB=%%i"
```
`-u` means convert to the unix filesystem namespace.
`-p` means convert a pathlist, i.e. a `;` delimited string, e.g. `C:\Some\Dir;C:\Another\Dir`

Because `cygpath.exe` is a POSIX program it understands the MSYS2/Cygwin mounts in `/etc/fstab` and does the right thing with them.

Alternatively, you might be able to use `mingw32-make` from the `m2w64-make` package instead. This is a native port of GNU make which knows nothing about the POSIX filesystem and contains some hacks to make it attempt to understand Windows paths a bit better. It doesn't work fantastically well in all cases, but if you try it and it fails, I can probably advise. FWIW, I never use `mingw32-make`, preferring `m2-make` and converting filepaths instead.

ocefpaf:
Thanks @gillins and @mingwandroid for the info. Now I understand much better what to expected when using `posix`.

However, I believe #129 is on the right track, so I am closing this one in favor of that.

gillins:
Yes thanks @mingwandroid - very informative! Makes a lot of sense.

PR #132, opened 2016-12-20
Fixes #131

### LINKED ISSUE #131: ImportError: No module named '_gdal_array' on py3
Hi, 

I don't know when it started but I think it is recent. It happens only on py3, on py2 there is no apparent problem:

```
/home/docs/checkouts/readthedocs.org/user_builds/oggm/conda/docs/lib/python3.5/site-packages/osgeo/gdalnumeric.py in <module>()
----> 1 from osgeo.gdal_array import *
      2 from numpy import *

/home/docs/checkouts/readthedocs.org/user_builds/oggm/conda/docs/lib/python3.5/site-packages/osgeo/gdal_array.py in <module>()
    166 
    167 import numpy
--> 168 import _gdal_array
    169 
    170 from . import gdalconst

ImportError: No module named '_gdal_array'
```

Thanks!

ocefpaf:
Not a fix yet... just a test to check if it fails everywhere.

See #131

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+* text=auto
+
+meta.yaml text eol=lf
+build.sh text eol=lf
+bld.bat text eol=crlf
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,4 +49,4 @@ script:
   - upload_or_check_non_existence ./recipe conda-forge --channel=main
 
   # inspect the prefix lengths of the built packages
-  - conda inspect prefix-lengths /Users/travis/miniconda3/conda-bld/osx-64/*.tar.bz2
+  #- conda inspect prefix-lengths /Users/travis/miniconda3/conda-bld/osx-64/*.tar.bz2
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,14 @@ install:
       conda install --yes --quiet conda-forge-build-setup
       source run_conda_forge_build_setup
 
+      # install conda-build 2.x to build with a long prefix
+      conda install --yes --quiet conda-build=2
+      conda info
+
 script:
   - conda build ./recipe
 
   - upload_or_check_non_existence ./recipe conda-forge --channel=main
+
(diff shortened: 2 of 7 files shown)


kmuehlbauer:
@ocefpaf It seems to fail with every py3k. Also this here at osgeo/gdal (https://ci.appveyor.com/project/OSGeo/gdal/build/1.0.721/job/bhll7kamd9serd6x#L10427).

ocefpaf:
Locally pinning to swig 3.0.8 fixed it. PR soon... Hold on.

kmuehlbauer:
Ahh, so the generation with a  swig > 3.0.8 is the culprit.

ocefpaf:
> Ahh, so the generation with a swig > 3.0.8 is the culprit.

Probably. I patched the code with the some changes from https://trac.osgeo.org/gdal/changeset/36845#file9 before I downgraded `swig`. So I am checking if we need both the patch and the `swig` downgrade or only the latter.

kmuehlbauer:
Just to understand this correctly: Aren't we [regenerating](https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/build.sh#L84) those files anyway?

ocefpaf:
> Just to understand this correctly: Aren't we [regenerating](https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/build.sh#L84) those files anyway?

The patch was for the `GNUmakefile`changes. Not sure if that file is auto-generated or not. Either way we do not need the patch.

kmuehlbauer:
@ocefpaf The Circle builds also have the short/long prefix issue. And I noticed that `swig` is pulled from `defaults`, because v3.0.10 is the first one in `conda-forge`.

ocefpaf:
Thanks for checling this @kmuehlbauer. The short prefix contaminating this build is probably from `swig`. But we don't really need the long prefix test so I commented that out.

Thanks @fmaussion for the bug report and @kmuehlbauer for the review!

PR #138, opened 2017-01-09
Support Windows Bash users

omad:
I use conda in Windows with the Bash shell,  so need environment variable activation 
shell scripts as well as batch scripts. I'm sure I'm not the only one.

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -111,3 +111,9 @@ if errorlevel 1 exit 1
 
 copy %RECIPE_DIR%\scripts\deactivate.bat %DEACTIVATE_DIR%\gdal-deactivate.bat
 if errorlevel 1 exit 1
+
+copy %RECIPE_DIR%\scripts\activate.sh %ACTIVATE_DIR%\gdal-activate.sh
+if errorlevel 1 exit 1
+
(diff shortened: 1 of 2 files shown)


ocefpaf:
I am not a Windows users to understand that change. So I will leave this one for @gillins to merge.

But you need to bump the [build number](https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/meta.yaml#L21) to get that change into a new package.

recipe/bld.bat
```diff
@@ -111,3 +111,9 @@ if errorlevel 1 exit 1
 
 copy %RECIPE_DIR%\scripts\deactivate.bat %DEACTIVATE_DIR%\gdal-deactivate.bat
 if errorlevel 1 exit 1
+
```
gillins: Can you put a comment in here about supporting `bash` on Windows?

gillins:
Hi @omad, thanks for the patch. I didn't know you could do this and I'll definitely be using `bash` from now on since the command prompt is so lame. Is there some `conda` documentation about using `bash` on Windows anywhere?

omad:
I haven't seen anything written in the conda documentation about unix like environments on Windows. But there have been discussions and commits on several repositories about it.

I've tried a few different options, but am currently Git Bash, which comes with the Git for Windows Installer, and https://conemu.github.io/ for a terminal.

PR #139, opened 2017-01-11
try using /MP option to cl to speed up build

gillins:
This PR adds the [/MP (Build with Multiple Processes) option](https://msdn.microsoft.com/en-us/library/bb385193.aspx) to be build flags for Windows which seems to speed up the build considerably for me locally. I've told it to use `%CPU_COUNT%` processors. Was a shame that there isn't an easier way of adding more compiler flags to the `gdal` build...

I also tried [`jom`](https://wiki.qt.io/Jom) but this didn't seem to help as the `gdal` makefile passes a number of source files to `cl` each time and `jom` didn't seem to be able to do much with this. /MP seemed to work better as each of these source files can be built on a different CPU.


ocefpaf:
The times are 3-5 minutes slower. In a build matrix with 14 items that is quite an improvement!
Thanks @gillins!

gillins:
>  3-5 minutes slower

You mean quicker? You had me worried..

ocefpaf:
> You mean quicker? You had me worried..

:smile: 

That is what I get for waking up early!

ocefpaf:
No need to wait for `Travis-CI`.

PR #150, opened 2017-03-24
hide batch command output

scw:
Currently, on deactivating an environment, this script outputs each step its executing. Hide these executions.


ocefpaf:
@scw same comments from https://github.com/conda-forge/proj.4-feedstock/pull/21#issuecomment-289025789

PR #153, opened 2017-04-12
Update gdal-activate.sh to work in Windows + Linux

omad:
Unfortunately Windows installs of GDAL place the data files in a different directory to Linux installed.

This patch checks before setting environment variables to see if the target directory exists.
It should fix GDAL environment variables for Windows bash users.

### DIFF
--- a/recipe/scripts/activate.sh
+++ b/recipe/scripts/activate.sh
@@ -6,12 +6,21 @@
 if [[ -n "$GDAL_DATA" ]]; then
     export _CONDA_SET_GDAL_DATA=$GDAL_DATA
 fi
-export GDAL_DATA=$CONDA_PREFIX/share/gdal
 
 if [[ -n "$GDAL_DRIVER_PATH" ]]; then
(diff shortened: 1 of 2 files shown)


recipe/scripts/activate.sh
```diff
@@ -6,12 +6,21 @@
...
+
+# On Linux GDAL_DATA is in $CONDA_PREFIX/share/gdal, but
+# Windows keeps it in $CONDA_PREFIX/Library/share/gdal
+if [ -d $CONDA_PREFIX/share/gdal ]; then
+    export GDAL_DATA=$CONDA_PREFIX/share/gdal
+    export GDAL_DRIVER_PATH=$CONDA_PREFIX/lib/gdalplugins
+elif [ -d $CONDA_PREFIX/Library/share/gdal ]; then
+    export GDAL_DATA=$CONDA_PREFIX/Library/share/gdal
+    export GDAL_DRIVER_PATH=$CONDA_PREFIX/Library/lib/gdalplugins
+fi
```
ocefpaf: I don't use Windows so I can't really comment here but these changes makes sense to me.
@gillins are you OK merging this?

PR #154, opened 2017-05-11
Update to 2.2.0

ocefpaf:
Dropping the regeneration of the python bindings in this release. It should not be needed. Let's hope the extra test can catch any possible problems with that.


ocefpaf:
Circle is timing out. I'll merge as is and then branch out some of the matrix items.

PR #156, opened 2017-05-18
Force a rebuild as new kealib version has broken gdal

gillins:
Get `error while loading shared libraries: libkea.so.1.4.6: cannot open shared object file: No such file or directory` when upgrade to latest `libkea` build. New library name is `libkea.so.1.4.7`. Not sure why `gdal` is linked to an exact version rather than just `1.4`....

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -20,7 +20,7 @@ source:
     - install_scripts.patch
 
 build:
-  number: 0
+  number: 1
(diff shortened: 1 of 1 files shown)


gillins:
Do not merge - working on creating extra links in the `libkea` feedstock instead....

gillins:
Closed in favour of: https://github.com/conda-forge/kealib-feedstock/pull/10

PR #157, opened 2017-06-02
Upgrade to hdf5 1.8.18

qwhelan:
A few CVEs were reported and fixed in hdf5 back in November 2016: http://blog.talosintelligence.com/2016/11/hdf5-vulns.html

The hdf5 release notes for that release are available here: https://support.hdfgroup.org/ftp/HDF5/current18/src/hdf5-1.8.18-RELEASE.txt

For additional info, please see: https://github.com/conda-forge/hdf5-feedstock/pull/68 and https://github.com/conda-forge/hdf5-feedstock/issues/71

PR #161, opened 2017-06-06
pin hdf5 and geos

ocefpaf:
Porting #157 and #160 to the `gdal2_PY36_linux` branch.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -20,7 +20,7 @@ source:
     - install_scripts.patch
 
 build:
-  number: 0
+  number: 1
   skip: True  # [not (py36 and linux)]
   features:
     - vc9  # [win and py27]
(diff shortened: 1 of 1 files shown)


ocefpaf:
The message below from `conda` helps a little:

```
  - kealib 1.4.7 -> hdf5 1.8.17|1.8.17.*
  - python 3.6*
```

@qwhelan we will need to navigate down all the dependencies to find what is going on. I believe the issue lies with the packages that depended on `libnetcdf`. I am working on it...

qwhelan:
@ocefpaf I think the dependency solver is having trouble as ```hdf5``` is not being built for ```win and py36``` (https://github.com/conda-forge/hdf5-feedstock/blob/master/recipe/meta.yaml#L29), but ```gdal``` is trying that combination.

ocefpaf:
That message is not very helpful. I (think) found the problem and I already re-start Windows and Linux. Waiting on https://travis-ci.org/conda-forge/kealib-feedstock/builds/240415587 to re-start Travis-CI

PR #162, opened 2017-06-06
Version the xerces-c dependency

microe:
The conda-forge xerces-c package is required, but it is possible that the default xerces-c will be brought in. I am hoping by versioning the package that the conda-forge xerces-c package will be picked up.


ocefpaf:
@microe see my comment on the issue. Unfortunately we cannot pin down to a build number (and we don't really need to pin the `xerces-c` version.)

The issue comes down to the mix of `defaults` and `conda-forge` packages. (And to avoid it you'll need to jump through a few hoops... Sorry.)

PR #163, opened 2017-06-19
Build python wrapper only using libgdal

ocefpaf:
# Do not merge this yet!

This PR is the second step to make a transition to the the `libgdal`+`gdal` package like we have in `defaults`. Note that `libgdal` will be the whole GDAL package **but** the python wrappers, while `gdal` will be the python wrapper only.

It is important to note that `gdal` will depend on `libgdal`, so this won't be a harmful change for `conda-forge` users because installing only `gdal` will install all the files the users expect. Meanwhile this change will make it easier for `defaults` users to migrate to `conda-forge` packages by avoiding the scenario where `gdal` (from `conda-forge`) + `libgdal` (from `defaults`)  are installed in the same env, not only overriding files but also bringing incompatible dependencies with them.

We need to merge https://github.com/conda-forge/staged-recipes/pull/3143 first!

If everything works this will eliminate the need for the extra branch in this feedstock.

The long term goal would be to merge the two feedstocks, `libdal` and `gdal`,  into one using `outputs` and `conda-build 3` new features. [Here](https://github.com/conda/conda-recipes/blob/master/xgboost/meta.yaml) is a nice example on how we can accomplish that.

Pinging @msarahan, @dharhas, and @conda-forge/gdal here to comment.


recipe/meta.yaml
```diff
@@ -12,79 +12,32 @@ source:
...
-    # https://lists.osgeo.org/pipermail/gdal-dev/2016-January/043443.html
-    - disable_jpeg12.patch  # [win]
-    - windowshdf5.patch  # [win]
-    - multiprocessor.patch  # [win]
     - install_scripts.patch
 
 build:
-  number: 2
-  skip: True  # [py36 and linux]
+  number: 3
```
ocefpaf: I bumped the build number here but I am also planning to move the previous build number to a "broken" channel to avoid confounding the conda solver.

PR #173, opened 2017-08-20
Sane numpy matrix 2.1.3

ocefpaf:


recipe/meta.yaml
```diff
@@ -82,7 +81,7 @@ requirements:
     - giflib 5.1.*  # [not win]
     - json-c  # [not win]
     - freexl
-    - openjpeg
+    - openjpeg >=2.1,<2.3.0
```
ocefpaf: This version needs to be built wtih `openjpeg 2.1` otherwise it could not find the header, but seems to work fine with later version. Maybe we should make this pin official...

gillins: Is this the same for all `gdal` versions or just 2.1.x?

ocefpaf: `2.1.x` needs it but I believe we can safely use that for all versions.

Issue #227, opened 2018-09-20
JPEG2000 support broken


cpaulik:
It seems that the latest gdal does not support jpeg2000 correctly. 

I get errors like:

```
$ gdalinfo /home/cpa/Downloads/L2A_T36JVT_20161028T073922_B02_10m.jp2              
ERROR 4: `/home/cpa/Downloads/L2A_T36JVT_20161028T073922_B02_10m.jp2' not recognized as a supported file format.
gdalinfo failed - unable to open '/home/cpa/Downloads/L2A_T36JVT_20161028T073922_B02_10m.jp2'.
```

Downgrading like this:

```
The following packages will be DOWNGRADED:

    gdal:    2.3.1-py27hb00a9d7_3 conda-forge --> 2.3.1-py27hb00a9d7_0 conda-forge
    libgdal: 2.3.1-hedf625b_3     conda-forge --> 2.3.1-ha6e4b24_0     conda-forge
    poppler: 0.67.0-h4d7e492_3    conda-forge --> 0.61.1-h4d7e492_4    conda-forge
```
fixes the problem. But build `2.3.1-py27hb00a9d7_0` is no longer available from `anaconda.org` so that does not help my colleagues.

There are some strange things here:

Running 

```
gdal-config --formats
```
does not list `openjpeg` but it seems to be linked correctly:

```
$ ldd libgdal.so | grep open
	libopenjp2.so.7 => /home/cpa/.pyenv/versions/miniconda2-4.3.30/envs/gdal_jp2000_test/lib/./././libopenjp2.so.7 (0x00007f733dccf000)
```

The test environment was created with the command `conda create -n gdal_jp2000_test -c conda-forge gdal python=2.7`

<br/>
Environment (<code>conda list</code>):
<details>

```
$ conda list
#
# Name                    Version                   Build  Channel
blas                      1.0                         mkl  
boost-cpp                 1.67.0               h3a22d5f_0    conda-forge
bzip2                     1.0.6                h470a237_2    conda-forge
ca-certificates           2018.8.24            ha4d7672_0    conda-forge
cairo                     1.14.12              he6fea26_5    conda-forge
certifi                   2018.8.24                py27_1    conda-forge
curl                      7.61.0               h93b3f91_2    conda-forge
expat                     2.2.5                hfc679d8_2    conda-forge
fontconfig                2.13.1               h65d0f4c_0    conda-forge
freetype                  2.9.1                h6debe1e_4    conda-forge
[...]

PR #228, opened 2018-09-24
WIP: attempt to determine why JPEG2000 support isn't working

gillins:
Checklist
* [ ] Used a fork of the feedstock to propose changes
* [ ] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

Do not merge. Attempt to debug #227 - `configure` works fine on my machine so something is different on the CI's. Also try to force the python tests to happen.


recipe/meta.yaml
```diff
@@ -17,12 +17,13 @@ source:
...
 
 build:
-  number: 3
+  number: 4
   skip: True  # [win and vc<14]
 
 requirements:
   build:
     - cmake  # [win]
+    - pkg-config  # [unix]
```
gillins: I'm now getting `OpenJPEG support:          yes` on the CI's with this which is a good start...

recipe/meta.yaml
```diff
@@ -186,6 +188,11 @@ outputs:
         - osgeo._gdalconst
         - osgeo._ogr
         - osgeo._osr
+      commands:
+        - $PYTHON run_tests.py  # [unix]
+        - %PYTHON% run_tests.py  # [win]
+        - $PYTHON extra_tests.py  # [unix]
+        - %PYTHON% extra_tests.py  # [win]
       about:
```
gillins: ... but I still can't work out how to get these tests to run. Any ideas?
```
can't open file 'run_tests.py': [Errno 2] No such file or directory
```

kmuehlbauer: @gillins It's `run_test.py` not `run_tests.py`

gillins: :man_facepalming: well spotted!

kmuehlbauer: @gillins Looking at `run_test.py`, `extra_tests.py` is imported there. So we do not need to call it specifically, or do I miss something?

kmuehlbauer: Not yet, still thinking...

recipe/meta.yaml
gillins: Not sure if I need to add `run_test.sh` to the test section here also to get it to run...

gillins:

recipe/install_python.bat
```diff
@@ -0,0 +1,13 @@
...
+
+%PYTHON% setup.py build_py
+if errorlevel 1 exit 1
+
+%PYTHON% setup.py build_scripts
+if errorlevel 1 exit 1
+
+%PYTHON% setup.py install
+if errorlevel 1 exit 1
+
```
ocefpaf: I guess we also need to remove `bld.bat`, right?

BTW, this is not how they did in AR, see https://github.com/AnacondaRecipes/gdal-feedstock/tree/master/recipe

Not sure if there is a reason or just style preference. Ping @msarahan here to comment on it.

recipe/meta.yaml
```diff
@@ -143,6 +144,8 @@ outputs:
     test:
       files:
         - test_data
+        - run_test_lib.sh  # [unix]
+        - run_test_lib.bat  # [win]
```
ocefpaf: My understanding is that the `run_test` would be run automatically without the need to list it. But not sure if they are run twice or only once with both packages installed. Either way I don't think we need this change. @msarahan probably knows more.

gillins: Hmm it doesn't seem to be here (or in #229). Well I can't find the outputs of these tests in the log (https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/run_test.sh#L7 and https://github.com/conda-forge/gdal-feedstock/blob/master/recipe/run_test.py#L13 should print something) though I'm sure I'm missing something.

BTW if these tests were run we would have caught the missing driver as `run_test.py` checks for it...

ocefpaf: Yep. You are, obviously, correct :grimacing: 

@msarahan I guess that it is a bug in conda-build, no? What is the recommend way to run tests for split packages in an integrated way? Or separated?

@gillins meanwhile let's merge #229 but let's keep working on this one to get the tests back up.

gillins:
ping @msarahan. Did this ever get resolved? I think there was some confusion about how/when tests are run for split packages....

Issue #244, opened 2018-12-04
New environments fail on Windows with Python 2.7


djhoese:
This is a continuation of #241. The recipe for GDAL 2.3.2 was updated in #243 to look for `tiff.dll` coming from the new builds of the `libtiff` package (instead of `libtiff.dll`). If I recall correctly gdal no longer supports python 2 on Windows (found it: https://github.com/conda-forge/gdal-feedstock/issues/241#issuecomment-443463792). This means that any gdal build for Windows Python 2.7 is still looking for the "old" `libtiff.dll` but is automatically finding the newer libtiff builds with `tiff.dll`. Resulting in:

```
(appveyor_debug_27) C:\Users\User>python -c "from osgeo import gdal, osr"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\User\Anaconda3\envs\appveyor_debug_27\lib\site-packages\osgeo\__init__.py", line 25, in <module>
    _gdal = swig_import_helper()
  File "C:\Users\User\Anaconda3\envs\appveyor_debug_27\lib\site-packages\osgeo\__init__.py", line 21, in swig_import_helper
    _mod = imp.load_module('_gdal', fp, pathname, description)
ImportError: DLL load failed: The specified module could not be found.
```

I could be wrong but I don't think the gdal package is pinned to a specific version/build of libtiff. Would it be possible to do this? Would you be willing to accept a PR for a new build of gdal 2.2.4 with the `tiff.dll` fixes backported? Or with `libtiff` version hardcoded to something earlier than the `tiff.dll` change?

Otherwise, I think the only option (considering the life cycle of python 2.7) would be to require uses to install a specific build of gdal and/or libtiff.

Thoughts?

Side note: I can't seem to get python 3.6 on Windows to work on appveyor with a brand new environment but my Windows VM is working fine. For some reason it isn't picking up the new gdal build.

ocefpaf:
> I could be wrong but I don't think the gdal package is pinned to a specific version/build of libtiff. 

That is probably the problem.

> Would you be willing to accept a PR for a new build of gdal 2.2.4 with the tiff.dll fixes backported? Or with libtiff version hardcoded to something earlier than the tiff.dll change?

Yes please :smile: 

(I was planning on doing that next week but if you can move this forward now that would be awesome.)

[...]

PR #245, opened 2018-12-04
Backport tiff.dll fix to v2.2.x

djhoese:
As discussed in #244 this backports the 2.3.x fix for the libtiff packages change from `libtiff.dll` to `tiff.dll`. This *should* fix python 2.7 environments on Windows.

Checklist
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)


djhoese:

.ci_support/linux_python2.7.yaml
```diff
@@ -13,31 +13,31 @@ docker_image:
 expat:
 - '2.2'
 gdal:
-- '2.2'
+- '2.3'
 geos:
-- 3.6.2
+- 3.7.0
```
ocefpaf: We may need to force this and others to older version to get vc9 working.
(Probably geos and hdf5, maybe proj.4 but I haven't checked.)

djhoese:
I think I need to force hdf5 to 1.10.3 so that netcdf4 and h5py can be used in this same environment on Windows. Thoughts @ocefpaf ?

.ci_support/linux_python2.7.yaml
djhoese: Yeah, looks like in my VM environment geos should be 3.6.2. I think hdf5 has to be 1.10.3 to get the most recent libnetcdf.

.ci_support/osx_python3.7.yaml
```diff
@@ -13,31 +13,31 @@ cxx_compiler:
 expat:
 - '2.2'
 gdal:
-- '2.2'
+- '2.3'
```
djhoese: @ocefpaf where/how is this getting set?

ocefpaf: These are the pins used in the build. Let me send a PR to conda-forge-pinning, where this comes from, adding an optional pinning for `vc9`, that means we may not build things for `vc9` on the new `gcc7`, but they can still work after a re-render. Hold on...

ocefpaf:
Let's get this in and I'll fix the vc9 build later.

PR #246, opened 2018-12-04
Add vc9

ocefpaf:
@djhoese if this does not build on `vc9` right now let's merge #245 and I'll figure out the necessary rebuilds/pins later.

### DIFF
--- a/.ci_support/win_c_compilervs2015cxx_compilervs2015python3.6vc14.yaml
+++ b/.ci_support/win_c_compilervs2015cxx_compilervs2015python3.6vc14.yaml
@@ -11,13 +11,13 @@ cxx_compiler:
 expat:
 - '2.2'
 gdal:
-- '2.3'
+- 2.2.4
 geos:
-- 3.7.0
+- 3.6.2
 hdf4:
 - '4.2'
 hdf5:
-- 1.10.4
+- 1.10.2
 icu:
 - '58'
 jpeg:
@@ -27,7 +27,7 @@ kealib:
 libkml:
 - '1.3'
 libnetcdf:
-- '4.6'
+- 4.4.1.1
 libpng:
 - 1.6.35
 libtiff:
@@ -87,7 +87,7 @@ pin_run_as_build:
   zlib:
     max_pin: x.x
 proj4:
-- 5.2.0
+- 4.9.3
 python:
 - '3.6'
 sqlite:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,13 +9,13 @@ environment:
     secure: tumuXLL8PU75WMnRDemRy02ruEq2RpNxeK3dz0MjFssnosPm2v4EFjfNB4PTotA1
 
   matrix:
-    - CONFIG: win_c_compilervs2008cxx_compilervs2008geos3.6.2hdf51.10.2kealib1.4.7libnetcdf4.4.1.1proj44.9.3python2.7vc9
+    - CONFIG: win_c_compilervs2008cxx_compilervs2008geos3.6.2hdf51.10.2libnetcdf4.4.1.1proj44.9.3python2.7vc9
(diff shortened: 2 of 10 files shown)


.ci_support/win_c_compilervs2015cxx_compilervs2015python3.6vc14.yaml
```diff
@@ -11,25 +11,25 @@ cxx_compiler:
...
 jpeg:
 - '9'
 kealib:
-- 1.4.9
+- 1.4.10
 libkml:
 - '1.3'
 libnetcdf:
-- '4.6'
+- 4.4.1.1
```
ocefpaf: This is not good :-/ only `vc9` should be changed.

djhoese: If I'm understanding you correctly, this is setting the vc9-only libnetcdf version on a python 3.6 recipe, right?

ocefpaf: This config file is `win_c_compilervs2015cxx_compilervs2015python3.6vc14.yaml`, which means it is `win and vc14`, so when I set the pre-selector for the new pins to be `win and vc<14` this one should not be modified. I expected changes only in the `win_c_compilervs2008cxx_compilervs2008python2.7vc9.yaml` config file.

ocefpaf:
Thanks to @isuruf I finally figured out the "right" way to create the `vc9` special case. See https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/160

Now we should start investigating the right combination of legacy Python and `vc9` versions of the packages for the pin.

ocefpaf:
No idea of what is the error on `vc9`:

```
(C:\bld\gdal-split_1544013616161\_test_env) C:\bld\gdal-split_1544013616161\test_tmp>gdal_grid --version 
(C:\bld\gdal-split_1544013616161\_test_env) C:\bld\gdal-split_1544013616161\test_tmp>IF -1073741515 NEQ 0 exit /B 1 
(C:\bld\gdal-split_1544013616161\_test_env) C:\bld\gdal-split_1544013616161\test_tmp>IF 1 NEQ 0 exit 1 
```

ocefpaf:
> @ocefpaf Could we look at the last successful Windows-py27 build for what packages to use?

That is the plan, but it is hard due to the huge number of dependencies here, not to mention that the feedstock used to be split in two at the time... I'll try to get to this over the weekend.

ocefpaf:
Even with https://github.com/conda-forge/kealib-feedstock/pull/27 this is a no go. I don't have the bandwidth to work in Legacy Python for `vc9` at the moment. I'm closing this but anyone with enough interest to move this forward please feel free to pick up.

kmuehlbauer:
@ocefpaf Thanks for trying!

djhoese:
@ocefpaf If I was to continue work on this, what would be my workflow? From what I understand I would have to:

1. Make a custom conda-forge-pinning package locally based on https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/160
2. rerender the gdal-feedstock recipe you started here
3. Check that only vc9 (Windows - Python 2.7) environments were changed in the recipe
4. commit and push to a new PR for this feedstock
5. Wait for builds to succeed
6. Have you review and merge it

Where did you leave off on trying to get this to work? What the pinning going fine, but the combination of packages was failing to compile?

My projects currently have a `libtiff.dll` copy to `tiff.dll` to get around these issues for now so I'll have to check if any of these environments are currently broken by default.

ocefpaf:
@djhoese mostly what you mentioned above but note that some of existing packages may be improperly pinned and/or need a rebuild. So, for that to work, we would need to rebuild the whole stack from the bottom up to "clear" whatever is the culprit in that `vc9` failure. That is an herculean job, specially on Windows. Feel free to bug me with questions though :smile:

Issue #217, opened 2018-07-26
Are jpeg12 and ECW support possible?


jacobadams-cc:
Issue:

Is there any possibility of including jpeg12 and ECW-read support in the conda-forge gdal feedstock? I've looked in the issue tracker and haven't seen either of these mentioned previously.

I'm running into issues with a jpeg-compeessed tiff from ArcGIS that is set to nbits=12, which seems to indicate conda-forge's gdal doesn't have jpeg12 support enabled as outlined in [https://trac.osgeo.org/gdal/wiki/TIFF12BitJPEG](url). Is it possible to enable this, or is there a history I've missed explaining why it's not?

While we're here, is there any chance of adding ECW read support as well? I realize this may not be feasible with all the licensing stuff.

My workaround for both of these is to use the gdal_translate.exe binary from a QGIS install to translate these files to a format supported by the conda-forge installation and then work from there. This still works, but it'd be nice to have direct support.

Thanks for maintaining this feedstock. Running an up-to-date version of gdal via conda-forge is so, so much easier than the other methods out there.
<br/>

(Omitting all the environment info as this is just a feature request. I just updated gdal to 2.3.1 from the conda-forge channel in Anaconda Navigator to confirm the behavior exists in the current build).

jacobadams-cc:
I'll take a stab at a PR; I've identified the relevant statements in the libgdal meta/build files.

Being new to conda development, do I need to test changes locally via conda build before submitting a PR, or is that what the conda-forge CI tools are intended to do? I can set up a windows build environment without that much headache, but osx and linux will take a while longer.

ocefpaf:
> do I need to test changes locally via conda build before submitting a PR, 

You don't have to do it but it is easier to fix errors that way.

>  I can set up a windows build environment without that much headache, but osx and linux will take a while longer.

OS X is usually harder but Linux you can use our docker image, even from Windows:

https://conda-forge.org/docs/testing.html#docker-tests

kevalshah90:
I am experiencing incompatibility issues with jpeg installing gdal library. I have tried pip and conda forge installations seperately for both the libraries, however, it seems like there is no jpeg support. 
 
![screen shot 2018-08-26 at 21 25 02](https://user-images.githubusercontent.com/4792353/44640528-0848ef00-a977-11e8-8308-c9535395e2cf.png)

ocefpaf:
You are probably hitting this conda issue: https://github.com/conda/conda/issues/7656

[...]

PR #255, opened 2019-01-02
WIP: Use internal jpeg sources

gillins:
Checklist
* [x] [Re-rendered]( https://conda-forge.org/docs/conda_smithy.html#how-to-re-render ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

xref: https://github.com/conda-forge/gdal-feedstock/issues/217

This PR attempts to revert back to the default GDAL behaviour which is to use the internal build of JPEG. This internal version appears to have diverged quite a bit from the upstream and includes support for 12 bit JPEGS (and 12 bit JPEG in TIFF). It isn't clear that upstream `jpeg` supports 12 bit and anyway GDAL doesn't allow you to have 12 bit and external `jpeg`.


gillins:

recipe/build.sh
```diff
@@ -31,7 +31,9 @@ bash configure --prefix=$PREFIX \
                --with-geotiff=$PREFIX \
                --with-hdf4=$PREFIX \
                --with-hdf5=$PREFIX \
-               --with-jpeg=$PREFIX \
+               --with-jpeg=internal \
+               --with-jpeg12 \
+               --with-hide-internal-symbols=yes \
```
gillins: I *think* this will hide jpeg symbols so there won't be clashes with `libjpeg`. Hopefully this doesn't cause other problems....

kmuehlbauer:
> This internal version appears to have diverged quite a bit from the upstream and includes support for 12 bit JPEGS (and 12 bit JPEG in TIFF).

@gillins Somehow 12bit JPEG in TIFF is not enabled (see [build](https://circleci.com/gh/conda-forge/gdal-feedstock/1084):
```
LIBJPEG support:           internal
12 bit JPEG:               yes
12 bit JPEG-in-TIFF:       no
```

gillins:
There is this message from the log:
```
configure: WARNING: Internal libjpeg12 has not the same ABI as libjpeg 8 bit. Disabling JPEG-in-TIFF 12 bit
```
Seems there is something wrong with the `configure` test? (see https://github.com/OSGeo/gdal/blob/master/gdal/configure.ac#L1901). Seems like it should work.

gillins:
Yes it does appear to be (https://github.com/conda-forge/libtiff-feedstock/blob/master/recipe/meta.yaml#L35). I'll try internal tiff. Sigh.

kmuehlbauer:
@gillins This is beginning to get crazy. Seems we also need libgeotiff=internal now.

mingwandroid:
Does anyone have contacts in the GDAL world who can explain why they are so keen to vendor things like this? Have they tried to upstream their changes and been rebuffed?

kmuehlbauer:
@mingwandroid AFAIK, this is to circumvent such problems on their side.

We can ask on gdal-dev list, how we should tackle this.

gillins:
@hobu I think you contribute to both `geotiff` and `gdal`? Do you have any thoughts on this?

hobu:
I think the 12-bit in jpeg might be coming from https://github.com/libjpeg-turbo/libjpeg-turbo rather than regular libjpeg, btw. I don't mess with this part of the codebase at all, but I'm happy to route if there's questions.

kmuehlbauer:
@hobu Thank you very much for this extensive insight into GDAL history and the possible path to avoid dragons out there.

rouault:
> I think the 12-bit in jpeg might be coming from https://github.com/libjpeg-turbo/libjpeg-turbo

No, it comes for IGJ libjpeg-6b (at least that's the version of the internal libjpeg12 in GDAL). I'm not sure if this was upstreamed in libjpeg-6b (I believe so, not sure at all, as IJG libjpeg was already in a deep dormant state). libjpeg-turbo might have just imported this 12-bit support, but no idea if it is working and working with GDAL in particular.

I believe that another reason for the internal versions of the libraries was that it made it slightly more straighforward to build GDAL on Windows, as packaging of open source software on Windows has historically be a 4th-class citizen.

But otherwise using the latest upstream versions of libtiff, libgeotiff, and libjpeg with GDAL is fine.

gillins:
Thanks @hobu and @rouault. I've tried adding `--with-jpeg12` and using the external libs but I still get the `configure: WARNING: Internal libjpeg12 has not the same ABI as libjpeg 8 bit. Disabling JPEG-in-TIFF 12 bit` message and JPEG12 in Tiff is disabled. 
`configure.ac` line 1901 seems to suggest that 12 bit JPEG in Tiff is only enabled if `JPEG_LIB_VERSION == 62` (presumeably this is the internal version). `nmake.opt` also states that "12bit libjpeg - use only with internal jpeg builds". 
[Wikipedia](https://en.wikipedia.org/wiki/Libjpeg) suggests that 12 bit was added to libjpeg in version 9a (we use 9b) but I have no idea how to test this.

I see that `libtiff` has a `jpeg12` option that we should enable, but I can't see that is going to make gdal configure happy.

I'm sure I'm missing something here - your help is very much appreciated!

gillins:
Closing as advice appears to be to use external `jpeg` (https://github.com/conda-forge/gdal-feedstock/pull/255#issuecomment-451235642) but `gdal` currently doesn't support external `jpeg` with 12 bit (https://github.com/conda-forge/libtiff-feedstock/pull/37#issuecomment-451727931). May reopen when it does support this.

Issue #253, opened 2018-12-20
2.3.3 and 2.4.0 Releases


hobu:
GDAL 2.3.3 was released today, and 2.4.0 will be released tomorrow. https://lists.osgeo.org/pipermail/gdal-dev/2018-December/049524.html

ocefpaf:
See PR #251 and the bot should get 2.4.0, no need to open issues unless there are actions we should take.

hobu:
GEOS 3.7.1 was released but the pinning is x.x.x and I think GDAL needs to be rerendered.

@conda-forge-admin, please rerender

conda-forge-linter:
Hi! This is the friendly automated conda-forge-webservice.

I just wanted to let you know that I rerendered the recipe in conda-forge/gdal-feedstock#256.

PR #256, opened 2019-01-17
MNT: rerender

conda-forge-linter:
Hi! This is the friendly automated conda-forge-webservice.

I've rerendered the recipe as instructed in #253.

Here's a checklist to do before merging.
- [x] Bump the build number if needed.


hobu:
Unfortunately this is now failing on OSX with the following:

```
old_placehold_placehold_placehold_placehold_placehold_placehold_place/lib/libxml2.la: No such file or directory
hold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib/libxml2.la: No such file or directory
libtool:   error: '/Users/distiller/miniconda3/conda-bld/gdal-d_placehold_place/lib/libxml2.la' is not a valid libtool archive
```

kmuehlbauer:
@hobu I checked with the gitter-channel. Let's try this recommended fix.

kmuehlbauer:
Just to not forget, we need to bump the build number before merge!

kmuehlbauer:
@isuruf Thanks! The `find $PREFIX/lib -name '*.la' -delete` did work! Could you elaborate a bit why that happened and if there is some documentation about it?

isuruf:
We started deleting .la files from packages and then conda-build started deleting .la files as well. But some old packages still have the .la packages. If B's .la depends on A's .la and when A is packaged .la file is deleted, but an old version of B is still used, then B's .la can't find A's .la. Solution is to delete all the .la files at the beginning

PR #270, opened 2019-03-17
Re render

ocefpaf:
Testing a strict channel build for macOS and Linux here to see if I can fix the broken state of the GIS on mac.

See https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/50/

### DIFF
--- a/.ci_support/linux_.yaml
+++ b/.ci_support/linux_.yaml
@@ -1,5 +1,7 @@
 c_compiler:
 - gcc
+c_compiler_version:
+- '7'
 channel_sources:
 - conda-forge,defaults
 channel_targets:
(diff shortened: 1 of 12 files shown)


ocefpaf:
This failure is odd:

```
conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform osx-64: {'clangxx_osx-64=4'}
```

@msarahan do you know why we cannot find `clangxx_osx-64=4` when using strict? I though that we should be able to install it from `defaults`. Or maybe I'm doing something wrong here.

Ping @CJ-Wright who was curios about the possible problems we would find when using `strict` in conda-forge.

ocefpaf:
@conda-forge/core this is passing now that we moved the conflicting compilers dependencies we had "vendored" to broken. Now `strict` is working as expected!

ocefpaf:
Closing this test. We should discuss this idea again before `conda 4.7` hit us hard and feedstocks start to break.

PR #285, opened 2019-05-11
Add CFITSIO to recipe for fits files 

csedwar:
cfitsio is in conda-forge added to meta.yaml, modified build.sh and set_bld_opts.bat to include what i think are the right options (more uncertain about windows options)

Checklist
* [X] Used a fork of the feedstock to propose changes
* [X] Ensured the license file is being packaged.


recipe/build.sh
```diff
@@ -33,6 +33,7 @@ bash configure --prefix=$PREFIX \
                --with-geos=$PREFIX/bin/geos-config \
                --with-geotiff=$PREFIX \
                --with-hdf4=$PREFIX \
+               --with-cfitsio=$PREFIX \
```
ocefpaf: Can you add a test for it?

gillins: > Can you add a test for it?

Just to clarify: In `run_test.py` from line 11.

csedwar: Should be done.  Had a typo "FTIS" vs "FITS" but I expect to pass.  Let me know what else you need

ocefpaf:
Looks like we have `cfitsio` on Windows, any reason to not add it here?

csedwar:
I thought it would be straightforward too.

I tried to compile a few times but could not figure out the appropriate configure options. Any suggestions?

ocefpaf:
> I thought it would be straightforward too. I tried to compile a few times but could not figure out the appropriate configure options. Any suggestions?

I would download the `cfitsio` Windows package, check where the `.lib` file lives, and then add it to `set_bld_opts.bat`. (You'll probably have to check the cmake options for `cfitsio`).

If that is too difficult you will have to skip the test on Windows so we can merge this.

csedwar:
Will try a few more things and see if I can get it to compile. 

I have some but not much windows experience so will see what I can do.

gillins:
Looks like adding the following to `set_bld_opts.bat` should work (some [comments](https://github.com/OSGeo/gdal/blob/master/gdal/nmake.opt#L534-L537) in `nmake.opt`):
```
FITS_PLUGIN=NO ^
FITS_INC_DIR=%LIBRARY_INC% ^
FITS_LIB="%LIBRARY_LIB%\cfitsio.lib" ^
```

recipe/set_bld_opts.bat
```diff
@@ -32,6 +32,9 @@ set BLD_OPTS=%WIN64% ^
     EXPAT_DIR=%LIBRARY_PREFIX% ^
     EXPAT_INCLUDE="-I%LIBRARY_INC%" ^
     EXPAT_LIB=%LIBRARY_LIB%\expat.lib ^
+		FITS_PLUGIN=NO ^
+		FITS_INC_DIR=%LIBRARY_INC% ^
+		FITS_LIB="%LIBRARY_LIB%\cfitsio.lib ^
```
gillins: Missing quote at end of this line (before the `^`). Also fix the indentation.

csedwar:
Sigh. Fixed.

gillins:
Thanks @csedwar !

PR #287, opened 2019-05-21
Rebuild for PROJ 6.1.0

snowman2:
@conda-forge-admin, please rerender.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -17,7 +17,7 @@ source:
     - patches/disable_jpeg12.patch  # [win]
 
 build:
-  number: 1
+  number: 2
   skip: True  # [win and vc<14]
 
 requirements:


snowman2:
Feel free to close if it is supposed to happen in the pins repo.

ocefpaf:
> Feel free to close if it is supposed to happen in the pins repo.

Proj4 is pinning to the exact version (`x.x.x`) so we need to change that in the global pinning. If you want to test it here it is fair game to add `6.1.0` in the `meta.yaml`. Note that we are pinning everything else to `gdal 2.4` b/c of `rasterio` and `fiona`, so this will only be used in case users are installing `gdal` alone in an environment.

snowman2:
I think changing the global pinning is the way to go. Thanks!

ocefpaf:
> I think changing the global pinning is the way to go. Thanks!

I just sent a PR for a migration to proj4 6.1.0. Let's close this one and wait for the bot!

snowman2:
👍

PR #289, opened 2019-05-24
set NETCDF_HAS_NC4=yes so netcdf can open files with hdf5 headers on Windows

gillins:
Checklist
* [x] Used a fork of the feedstock to propose changes
* [x] Bumped the build number (if the version is unchanged)
* [x] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [x] Ensured the license file is being packaged.

Fixes https://github.com/conda-forge/gdal-feedstock/issues/279.

PR #290, opened 2019-05-24
some backports and add branch for v2.4

ocefpaf:
@snowman2 #289 fixes for `gdal 3.0` but you'll need it for `gdal 2.4` from this PR.


snowman2:
Good catch, thanks!

PR #295, opened 2019-06-17
Proj52

snowman2:
@conda-forge-admin please rerender.

### DIFF
--- a/.ci_support/linux_.yaml
+++ b/.ci_support/linux_.yaml
@@ -20,8 +20,6 @@ gdal:
 - '2.4'
 geos:
 - 3.7.1
-geotiff:
-- 1.5.1
 giflib:
 - 5.1.7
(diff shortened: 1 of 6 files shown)


ocefpaf:
Not sure what is going on with the Windows builds there but building packages with older versions is never easy :-/

snowman2:
Yeah, the GDAL stack is especially fun 🙃. This is not a blocker at the moment, so I wouldn't loose any sleep over it. Thanks for updating the pins BTW.

gillins:
I've seen this error before - not sure how it happens - but workaround is to remove the duplicate definition. Seems to work ok now.

ocefpaf:
Awesome! Thanks @gillins!

snowman2:
Nice, thanks @gillins!

PR #306, opened 2019-07-12
rebuild without cfitsio

ocefpaf:
Should with the packages that still rely on the old `proj4 <6`, like `cartopy` and `rasterio`, but got hit by the `cfitsio` ABI breakage.


ocefpaf:
Merging to get Windows fixed and I'll deal with Linux and macOS later.

PR #310, opened 2019-07-19
Rebuild

ocefpaf:
Same as #307 for the old `proj4 5` branch.

Should fix https://github.com/conda-forge/cartopy-feedstock/issues/63 and https://github.com/conda-forge/gdal-feedstock/issues/279

PR #311, opened 2019-07-20
update to 2.4.2

ocefpaf:
Build 2.4.2 with proj4 5.2.0.


ocefpaf:
Merging this without windows. We have version `2.4.1` with `proj4 5.2.0` from #310 and it is not worth fixing this for old builds.

PR #330, opened 2019-08-23
Build with userfaultfd support

### LINKED ISSUE #323: Build with userfaultfd support
Issue:

GDAL is currently not being built with userfaultfd support. We rely on conda for teaching GDAL and related GIS tools with a jupyterhub setup; and currently can't use the new vsis3 features for NETCDF datasets (e.g, from JPL's ARIA system/ ASF SAR archive). Looking through the azure pipelines and https://github.com/OSGeo/gdal/issues/1328 - its probably because of "userfaultfd" presence and usability are reported as "no" by configure - https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=55722. 

Would it be possible to enable "userfaultfd" support for conda-forge builds?

<br/>
Environment:

Standard linux x86_64 with gdal 3.0.1 installed via conda_forge (no other packages explicitly installed).

```
$ conda info
     active environment : gdal3
    active env location : /scratch/agram/python/miniconda3/envs/gdal3
            shell level : 2
       user config file : /home/agram/.condarc
 populated config files : /home/agram/.condarc
          conda version : 4.7.10
    conda-build version : not installed
         python version : 3.7.3.final.0
       virtual packages : __cuda=10.1
       base environment : /scratch/agram/python/miniconda3  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /scratch/agram/python/miniconda3/pkgs
                          /home/agram/.conda/pkgs
       envs directories : /scratch/agram/python/miniconda3/envs
                          /home/agram/.conda/envs
               platform : linux-64
             user-agent : conda/4.7.10 requests/2.22.0 CPython/3.7.3 Linux/3.10.0-957.10.1.el7.x86_64 centos/7.6.1810 glibc/2.17
                UID:GID : 42464:1001
             netrc file : /home/agram/.netrc
           offline mode : False
```

piyushrpt:
Checklist

Fixes #323 

1. Solution was to drop the userfaultfd.h header for compilation and then remove it. 
2. Had to define "__NR_userfaultfd" variable in the headerfile as this does not appear to be defined in the x86_64 environment on centos 6. Can possibly move that to a patch to cpl_userfaultfd.h.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -19,6 +19,8 @@ fi
 # See https://github.com/AnacondaRecipes/aggregate/pull/103
 if [[ $target_platform =~ linux.* ]]; then
   export LDFLAGS="${LDFLAGS} -Wl,-rpath-link,${PREFIX}/lib"
+  mkdir -p ${PREFIX}/include/linux
+  cp ${RECIPE_DIR}/userfaultfd.h ${PREFIX}/include/linux/userfaultfd.h
 fi
 
 # `--without-pam` was removed.
@@ -60,3 +62,7 @@ bash configure --prefix=$PREFIX \
                $OPTS
 
 make -j $CPU_COUNT ${VERBOSE_AT}
+
+if [[ $target_platform =~ linux.* ]]; then
+  rm ${PREFIX}/include/linux/userfaultfd.h
+fi
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -16,7 +16,7 @@ source:
     - patches/disable_jpeg12.patch  # [win]
 
 build:
-  number: 7
+  number: 8
   skip: True  # [win and vc<14]
 
 requirements:
--- a/recipe/userfaultfd.h
+++ b/recipe/userfaultfd.h
@@ -1,0 +1,237 @@
+/* SPDX-License-Identifier: GPL-2.0 WITH Linux-syscall-note */
+/*
+ *  include/linux/userfaultfd.h
+ *
+ *  Copyright (C) 2007  Davide Libenzi <davidel@xmailserver.org>
+ *  Copyright (C) 2015  Red Hat, Inc.
+ *
+ */
+
+#ifndef _LINUX_USERFAULTFD_H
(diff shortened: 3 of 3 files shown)


piyushrpt:
This should be ready. Any feedback / suggestions will be helpful.

piyushrpt:
Pinging @msarahan @ocefpaf @mingwandroid @kmuehlbauer @gillins for feedback / next steps

gillins:
Is it possible to add a test that this functionality is being enabled?

Also, would it be possible to use `userfaultfd.h` from centOS (via `yum_requirements.txt`) rather than shipping it with this recipe?

mingwandroid:
We should use cdt packages (as well as yum_requirements.txt) for this.

piyushrpt:
Centos 6 does not include this header. Newer kernels support it and other packages like qemu get around this limitation by packaging the header. Any suggestions on which yum package to use - maybe a custom one that I'm not aware of.

gillins:
> Centos 6 does not include this header.

OK so we won't be able to run any tests on the CI's based on this functionality? Not sure we have a precedence on this @ocefpaf @isuruf ?

piyushrpt:
Here is the link to the conversation on gdal-dev list: 
http://osgeo-org.1560.x6.nabble.com/gdal-dev-Packaging-linux-userfaultfd-for-conda-td5413625.html

As long as the binary gets built, it should work. I looked at kernel versions of the CI systems under use and they should handle this functionality. I will look into adding a unit test - probably one from the gdal test suite with a publicly hosted dataset.

ocefpaf:
> OK so we won't be able to run any tests on the CI's based on this functionality? Not sure we have a precedence on this @ocefpaf @isuruf ?

No ideal but not a show stopper. I don't think we should wait for a new docker image, let's give this a try.

PR #332, opened 2019-08-24
DOT NOT MERGE THIS ONE :: IT IS JUST AN EXAMPLE

ocefpaf:
@jomey as soon as you left the building I got what was wrong with the build. This should work. Just copy the changed parts here in your setup.

BTW, the easiest way to install a deve release in a conda environment is to create an env with gdal and then force remove it, and install the artifact you'll get here. Something like:

```
conda create --name TEST python=3.7 gdal libgdal
conda activate TEST
conda uninstall gdal libgdal --force-remove
conda install <name of the artifact you'll get on your build>
```

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -2,6 +2,9 @@
 
 set -e # Abort on error.
 
+cd gdal
+./autogen.sh
+
 # recommended in https://gitter.im/conda-forge/conda-forge.github.io?at=5c40da7f95e17b45256960ce
 find $PREFIX/lib -name '*.la' -delete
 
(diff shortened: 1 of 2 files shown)


msarahan:
Conda install --only-deps

Might be good instead of the force removal

ocefpaf:
> conda install --only-deps

Wait! How did I not know that already! Thanks @msarahan!

jomey:
Thank you @ocefpaf for figuring this out! 💯 👍

ocefpaf:
> Thank you @ocefpaf for figuring this out!

Note that there is a legit error in the compilation. Not sure how to fix that. It may be a problem with latest master and/or a combination with our package versions. I'll take a closer look soon.

recipe/build.sh
```diff
@@ -2,6 +2,9 @@
 
 set -e # Abort on error.
 
+cd gdal
+./autogen.sh
```
ocefpaf: @jomey these are the two lines that you want to copy.

PR #345, opened 2019-11-05
Powershell activation scripts

peterdhansen:
These are some simple scripts that just mirror the .bat files but in powershell.

So I am not sure what else might need to be fixed, but I ran into this issue and wanted to get it fixed somewhere.

Turns out conda in powershell only uses .ps1 files in /etc/activate.d/ and /etc/deactivate.d/ so powershell on windows does not get these environment variables.


ocefpaf:
Please bump the build number to create a new package.

msarahan:
> Turns out conda in powershell only uses .ps1 files in /etc/activate.d/ and /etc/deactivate.d/ so powershell on windows does not get these environment variables.

That's odd...  I thought it had some magic to grab stuff from the existing .bat files.  I'll check on that.

msarahan:
Seems like this would be the place: https://github.com/conda/conda/blob/master/conda/shell/condabin/Conda.psm1#L69-L106

but nothing there - for .bat or for .ps1.  I'm probably missing something, though.

ocefpaf:
Thanks Mike, merging this for now.

peterdhansen:
> Seems like this would be the place: https://github.com/conda/conda/blob/master/conda/shell/condabin/Conda.psm1#L69-L106
> 
> but nothing there - for .bat or for .ps1. I'm probably missing something, though.

Looks like the design is for only one extension per shell type, so each feedstock entry should really implement each shells activate/deactivate. 
https://github.com/conda/conda/blob/master/conda/activate.py#L1026

There is discussion on how to make these shell-agnostic

https://github.com/conda/conda/issues/6820

PR #372, opened 2020-03-29
Bump to build no 3 and rerender

hobu:
Rerendering to try to fix #371


kmuehlbauer:
@hobu There is again a soname change in libnetcdf (4.7.4 -> libnetcdf.so.18, instead of libnetcdf.so.15). 

The :hankey: is hitting the fan... ping @ocefpaf

ocefpaf:
I removed the latest `libnetcdf` for now. I'll add a stricter pin for it and pu a migrator to solve this. I asked upstream about this kind of breakages in the past and they mentioned they do no promise SEMVER, so only a stricter pin to libnetcdf will safeguard us from futures changes like this one.

ocefpaf:
BTW, closing this b/c we need to rebuild some other dependencies before getting to `gdal`. Thanks @hobu!

PR #404, opened 2020-07-21
CFEP-18: Remove static libraries from main build

xhochy:
This removes the static libraries from the main package according to [CFEP-18](https://github.com/conda-forge/cfep/pull/34). This might break some downstream builds but this is intended and we should fix the downstream builds to use the shared libraries before adding an output here.

Checklist
* [x] Used a [fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -12,7 +12,7 @@ source:
     - disable_jpeg12.patch  # [win]
 
 build:
-  number: 1
+  number: 2
 
 requirements:
   build:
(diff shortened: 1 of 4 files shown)


recipe/meta.yaml
```diff
@@ -132,17 +132,6 @@ outputs:
...
-        - gdal_grid --version
-        - gdal_rasterize --version
-        - gdal_translate --version
-        - gdaladdo --version
-        - gdalenhance --version
-        - gdalwarp --version
-        - gdalinfo --formats
-        - gdalinfo http://thredds.nersc.no/thredds/dodsC/greenpath/Model/topaz  # [not win]
-        - run_test.bat  # [win]
-        - bash run_test.sh  # [not win]
```
ocefpaf, line 145: @xhochy what is the reason to move these? I have to confess that I prefer them here so we can fix them in a single place rather than in two different files.

xhochy: If there is a `run_test` script, these commands were not executed at all. I moved them into the scripts to be run.

ocefpaf: Ouch! I had no idea and never checked. That is a bug IMO but let's get this in as-is and maybe we can fix it upstream.

PR #566, opened 2021-11-10
Cross compile linux aarch and ppc on linux 64

akrherz:
Checklist
*  Reset the build number to `0` (if the version changed)

~~I have no idea if I done this right, but the rerender seems to have done what I thought was needed?  I throw myself at the benevolence of the conda-forge experts to kindly educate me if this is the wrong approach...~~

PR morphed into cross compiling aarch and ppc on linux 64 to gain ~5-6x speedup.

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -15,11 +15,11 @@ jobs:
       linux_aarch64_:
         CONFIG: linux_aarch64_
         UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-aarch64
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-comp7
       linux_ppc64le_:
         CONFIG: linux_ppc64le_
         UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-ppc64le
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-comp7
   timeoutInMinutes: 360
 
   steps:
--- a/.ci_support/linux_aarch64_.yaml
+++ b/.ci_support/linux_aarch64_.yaml
@@ -21,7 +21,7 @@ cxx_compiler:
 cxx_compiler_version:
 - '9'
 docker_image:
-- quay.io/condaforge/linux-anvil-aarch64
+- quay.io/condaforge/linux-anvil-comp7
 expat:
 - '2'
 geos:
--- a/.ci_support/linux_ppc64le_.yaml
+++ b/.ci_support/linux_ppc64le_.yaml
@@ -17,7 +17,7 @@ cxx_compiler:
 cxx_compiler_version:
 - '9'
 docker_image:
-- quay.io/condaforge/linux-anvil-ppc64le
+- quay.io/condaforge/linux-anvil-comp7
 expat:
 - '2'
 geos:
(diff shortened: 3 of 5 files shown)


xylar:
@akrherz, each of these builds will produce a redundant `libgdal` with the same hash.  Isn't that going to be dangerous?

xylar:
@ocefpaf, what do you think about this approach?  Is there someone else from core we should check with?

akrherz:
> Isn't that going to be dangerous?

Yes, we would not want that.  I was blinding thrashing here and am researching this more now.  Here's what I see from the linux x64 build that finished

Label | gdal | libgdal
--- | --- | ---
linux linux_64_numpy1.18python3.7.____cpython | gdal-3.4.0-py37h7b489ff_2 | libgdal-3.4.0-hc0e84f1_2
linux linux_64_numpy1.18python3.8.____cpython |gdal-3.4.0-py38hcf2042a_2 | libgdal-3.4.0-hc0e84f1_2
linux linux_64_numpy1.19python3.9.____cpython | gdal-3.4.0-py39h0494519_2 | libgdal-3.4.0-hc0e84f1_2
linux linux_64_numpy1.21python3.10.____cpython | gdal-3.4.0-py310h222ae0d_2 | libgdal-3.4.0-hc0e84f1_2

xylar:
It's good that all the redundant `libgdal` builds have the same hash but we would need to somehow designate only one of the jobs to do the actual upload.  I don't know if that's possible.

akrherz:
@xylar what about a `skip` on the outputs for anything other than python 3.10 ?  Too brittle?

xylar:
@akrherz, that unfortunately won't work because each `gdal` build requires the associated `libgdal` build.  My suggestion from earlier was that we unarchive the `libgdal` feedstock and only build the `gdal` packages here.  That would be similar to how `esmf` and `esmpy` works.

ocefpaf:
> @akrherz, that unfortunately won't work because each `gdal` build requires the associated `libgdal` build. My suggestion from earlier was that we unarchive the `libgdal` feedstock and only build the `gdal` packages here. That would be similar to how `esmf` and `esmpy` works.

I'm not a big fan of that but at this point you two are more capable of deciding the future of this feedstock than me :-)

If you believe that will help we will need to adjust the dependency graph, to get the bot to work properly, and unarchive libgdal.

akrherz:
well, nothing is being saved on the build times here :sob:

xylar:
> well, nothing is being saved on the build times here sob

I think the problem is that building `libgdal` takes 96% of the time and building the 4 python packages takes the remaining 4 percent.  These are probably exaggerations but the point is that `libgdal` could take the full 6 hours on its own.  In which case we haven't won anything significant by splitting up the package. 

@ocefpaf, thanks for the input.  I don't think there's good evidence that we want to go that route and your arguments for not doing so are convincing.

Are there any options for requesting longer azure timeouts for this package?

akrherz:
Has anybody tried cross compiling to see how slow that is?  I have redid this PR to do that now to see how badly it goes up in flames.

xylar:
> Has anybody tried cross compiling to see how slow that is?

I'm not quite sure how things work on Azure.  I think it might always be cross-compiling but I haven't seen a clear explanation.  In any case, everything seems about equally slow either way.  It would be hard for me to imaging that cross-compiling would speed things up.

akrherz:
@xylar It appears to be 5-6x faster, augh, sigh, horray?

xylar:
@akrherz, hmm, that's interesting!

xylar:
@akrherz, well, I was clearly wrong.  I'm glad to know this for future reference!

Do you want to rename this PR and update the description so we can merged it?  And probably make the same change for `3.3.x`?

akrherz:
@xylar done and done and done.

xylar:
Crazy, I can hardly believe it!

Issue #585, opened 2021-12-01
Create GDAL 3.3/3.4+ proj 8.2 builds against icu68


akrherz:
I think there is a fair amount of dependency log jam within conda-forge due to the stuck icu69 migration for qt,  conda-forge/qt-feedstock#207.  It appears significant work is happening within the qt repo and I am unsure of how long it may take.

My 2 cents is that we could help this situation out greatly by creating new GDAL 3.3 and 3.4 builds with proj 8.2 and against icu 68.  I am willing to do that work, but am unsure of the "blessed" approach.  My ignorant idea is to:

1. PR a reversal of the icu69 migration with build number bump.
2. Merge that
3. Migrate icu69 again with build bump and merge that

akrherz:
As I debug other repos with dep troubles, I find this puzzling combo, which may be helped with this change...

```
$ mamba create -n dev python=3.10 proj=8.1.1 gdal=3.3 geos=3.10.0 icu=68
...
  - package gdal-3.3.3-py37h7b489ff_1 requires python_abi 3.7.* *_cp37m, but none of the providers can be installed
```

akrherz:
Well, maybe not as we would also need to temp revert proj 8.2.0 migration :)

akrherz:
The above maybe just a quirk of the mamba solver, indeed

libgdal-3.3.3-hf5a45a9_1.tar.bz2 -> `icu >=68.2,<69.0a0` + `proj >=8.1.1,<8.1.2.0a0`

So this change would not help those still stuck on proj 8.1.1, but proj 8.2.0 migrations would work.

gillins:
How many projects are held up by this? If it's only `qgis` I'd say let's wait as the current `qgis` version being built is so old I suspect no one is using it...

akrherz:
> How many projects are held up by this?

Fair question and likely not enough to fuss about, but am trying to get migrations finished out as I can and am seeing the icu69 issue about everywhere I poke around these days.

xylar:
@akrherz, I think this will be a huge headache and likely isn't worth the trouble. 

Regarding your comment https://github.com/conda-forge/gdal-feedstock/issues/585#issuecomment-984089768, I think that strange failure message just means that there isn't a solution.  As you know, the `mamba` output isn't always very helpful or intuitive on why.

akrherz:
> I think this will be a huge headache

[...]

PR #590, opened 2021-12-04
create intermediate build for qt compatability

akrherz:
Checklist
*  Reset the build number to `0` (if the version changed)
* :heavy_minus_sign:  [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)

see #585 

This is an attempt to create an intermediate build of gdal 3.4.0 with a combination of dependencies that should play nice with the most recent build of `qt`.  As an initial sanity test, this resolves within mamba.

$ mamba create -n dev python=3.10 libpq=13.1 icu=68 expat=2.2 qt

I purposefully did not rerender here.  We will want to:

- [x] carefully inspect the CI reported build dependencies to see if my hard coding here worked.

Credit for this suggestion goes to @xylar, thanks!  Blame for code execution and subsequent breakages goes to me!


akrherz:
Well, that quickly went up in dependency flames

```
   Encountered problems while solving:
     - nothing provides libcurl 7.79.0 h2574ce0_0 needed by curl-7.79.0-hea6ffbf_0
```

and debugging another step

```
$ mamba create -n dev python=3.10 libpq=13.1 icu=68 expat=2.2 qt libcurl=7.79
- package libcurl-7.79.1-h2574ce0_0 requires krb5 >=1.19.2,<1.20.0a0
```

Ah, libpq=13.1 is too tight and not necessary.  Attempt 2 ahoy...

akrherz:
Hmmm again

openssl1 build
```
   Encountered problems while solving:
     - nothing provides requested openssl 1.1.1
```

openssl3 build
```
   Encountered problems while solving:
     - package libpq-13.1-hfd2b0eb_0 requires openssl >=1.1.1i,<1.1.2a
```

:cry: there is no libpq=13 + openssl3 build, but that does not explain the openssl=1 failure....

akrherz:
The [postgresql 13.5 release notes](https://www.postgresql.org/docs/release/13.5/) indicate openssl3 support was added with that point release, so I suppose I could submit a PR there :)

xylar:
@akrherz, I did a local build with openssl1 on linux-64 and everything worked fine.  I don't understand why `mamba` is having trouble in CI but not when I try building.  Going to see if restarting CI happens to help.

xylar:
BTW, I don't think editing the files in `.ci_support` is the "right" way to do this.  Instead, we would probably want to specify versions that differ from the pins in the `meta.yaml` (both in host and run).  We could use a `recipe/conda_build_config.yaml` in cases were we need to specify a matrix.

xylar:
Hmm, obviously the solver issue is persistent.  But I can't reproduce it locally so I don't think I can help further at least for now.

xylar:
I wouldn't worry about openssl3 for now.  We could just remove it for this build.

akrherz:
Thanks again @xylar , what do you think of the latest commit?

xylar:
@akrherz, it looks like a good approach!  I think you may still need a constraint on `expat` but I could be wrong.

akrherz:
> I think you may still need a constraint on `expat`

You think it needs to be `>=2.2,<2.4` or ?

xylar:
Hmm, maybe let it be for now.  What I'm seeing is:

```
$ mamba search --info qt
...
  - expat >=2.2.10,<3.0.0a0
...

$ mamba search --info libgdal=3.4.0
...
   - expat >=2.4.1,<3.0a0
...
```
Those should be compatible.  The pin for `qt` is `expat=2.2`, but I guess that gets rendered as `expat >=2.2,<3`?

xylar:
In terms of debugging the solver issue, I don't have a better suggestion at this point than to just start commenting out dependencies and rerendering until the that initial host solve works.

xylar:
If you have docker installed locally and can reproduce the issue, that would likely be much faster than doing it here.

akrherz:
> If you have docker installed locally and can reproduce the issue, that would likely be much faster than doing it here.

I am having trouble getting it to work with podman :(  I am reproducing the dep issue though.

akrherz:
And gasp, changing "openssl=1.1.1" to "openssl=1.1.1l"  and mamba works :(

xylar:
> And gasp, changing "openssl=1.1.1" to "openssl=1.1.1l" and mamba works :(

Interesting.  I think the conda solver (and maybe also the mamba solver) has problems sometimes with deciding whether a given version number is legit or a release candidate that it should ignore.  This might be such a situation.

xylar:
Maybe use the pin that `gdal` was rendering with:
```
'openssl >=1.1.1l,<1.1.2a'
```

akrherz:
So my understanding is that not having a `openssl=3` build here is not of concern for the `qt` compatibility as `qt` has a direct openssl requirement and was not migrated yet.  So feedstocks with an openssl3 requirement + qt are stuck anyway, irregardless of gdal.  This change should help feedstocks that have no-openssl requirement, but have a qt + gdal requirement.  I do surely hope.

Anxiously awaiting to review build output requirements output :)

akrherz:
Well, the reported recipes at the end of the build neither report icu nor libpq, sigh.  From my reading of the build and compile logs, I believe this did what we intended...

xylar (approved):
I'm good with this as long as there's a PR waiting in the wings to reverse these changes and bump to build 12.

akrherz:
> I'm good with this as long as there's a PR waiting in the wings to reverse these changes and bump to build 12.

Thanks, Will do!

PR #603, opened 2022-01-23
[ci skip] [skip ci] [cf admin skip] ***NO_CI*** drop 3.3 migration branch

akrherz:
Given the build failure with newer `poppler` in #599, perhaps it is time to let go of the `3.3.x` bot migration branch.


gillins (approved):

xylar (approved):
I think this uses a lot of CI and human time, so if we don't really, really need it, we should drop it.

PR #619, opened 2022-04-01
run_test.sh: remove DODS testing

rouault:
I'm unclear why the recipe tests this particular driver... Anyway I'm
about to either remove that driver upstream (cf
https://github.com/OSGeo/gdal/issues/5173), or perhaps less radical
option, just require to have a DODS_RASTER: prefix at the beginning of the
connection string. In any case, if not changing this recipee, this would
break the GDAL upstream CI task that uses this recipee to build a Conda package from master.

(I don't think this justify bumping the version number?)

### DIFF
--- a/recipe/run_test.sh
+++ b/recipe/run_test.sh
@@ -49,9 +49,5 @@ gdalenhance --version
 gdalwarp --version
 gdalinfo --formats
 
-# allow the DODS test to run for now
-export GDAL_ENABLE_DEPRECATED_DRIVER_DODS=YES
-gdalinfo http://thredds.nersc.no/thredds/dodsC/greenpath/Model/topaz
-
 test -f ${PREFIX}/lib/libgdal${SHLIB_EXT}
 test ! -f ${PREFIX}/lib/libgdal.a


gillins:
I'm ok with this. @ocefpaf do you have any recollection why this was added?

@rouault I think you still need to bump the build number otherwise new packages won't be created.

rouault:
> I think you still need to bump the build number otherwise new packages won't be created.

For my immediate needs, I don't need new packages to be created. But I'm fine to do it if you think it's best

gillins:
Ah ok, I see this is for your test builds where you are testing your trunk - no problem. Let us know if you do end up removing that driver...

ocefpaf:
I probably added it for completeness' sake when we added that driver. It is ok to remove. (Sorry I'm late here, slowly catching up with the notifications.)

gillins:
Thanks @ocefpaf no need to apologize!

Actually @rouault do you have a version of our build scripts that with the new `cmake` build system? Would save us a lot of work with the upcoming changeover if we didn't have to start from scratch...

PR #637, opened 2022-07-01
Handle spaces in %CONDA_PREFIX%

j9ac9k:
activate.bat in its current form does not handle the case where there is
a space in %CONDA_PREFIX%, and thus it attempts to set %GDAL_DRIVER_PATH% to
a non-existent directory.

This PR allows for the proper escaping of spaces in %CONDA_PREFIX% and %GDAL_DRIVER_PATH%

Checklist

This issue seems to be most common in cases where a user has a space in their windows user name, and conda is installed to the user profile in `%USERPROFILE%/miniconda`

This issue can be replicated in the following manner

```doscon
mkdir "%USERPROFILE%\some space"
conda create -p "%USERPROFILE%\some space" -y
conda activate "%USERPROFILE%\some space"
conda install -c conda-forge pdal -y
...
Proceed ([y]/n)? y

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
The system cannot find the path specified.

C:\Users\ogi\some space>set "GDAL_DRIVER_PATH="
The system cannot find the path specified.

C:\Users\ogi\some space>set "PDAL_DRIVER_PATH="

C:\Users\ogi\some space>rem proj-data is installed because its license was copied over
```

A similar PR was submitted to the PDAL feedstock as well.

The powershell activate script does not appear to be affected in the same manner.


xylar:
The CI failures seem to be unrelated to GDAL and instead are from:
https://github.com/conda-forge/numcodecs-feedstock/issues/87

recipe/scripts/activate.bat
```diff
@@ -2,17 +2,17 @@
...
 
 @if defined GDAL_DRIVER_PATH (
-    set "_CONDA_SET_GDAL_DRIVER_PATH=%GDAL_DRIVER_PATH%"
+    set _CONDA_SET_GDAL_DRIVER_PATH="%GDAL_DRIVER_PATH%"
 )
 
 @REM Support plugins if the plugin directory exists
 @REM i.e if it has been manually created by the user
-@set "GDAL_DRIVER_PATH=%CONDA_PREFIX%\Library\lib\gdalplugins"
-@if not exist %GDAL_DRIVER_PATH% (
```
gillins, line 16: I'm no expert but just checking https://ss64.com/nt/set.html (see the section "Values with Spaces - using Double Quotes") it looks like the quotes will be stored unless you quote the whole thing. ie:
```
set _CONDA_SET_GDAL_DATA="%GDAL_DATA%"
```
Will store the quoted value of GDAL_DATA, but:
```
set "_CONDA_SET_GDAL_DATA=%GDAL_DATA%"
```
(the original) will store the value of GDAL_DATA without quotes (but includes spaces). 
I'm happy to accept there is a problem with `activate.bat`, I'm just thinking that the problem you are seeing maybe elsewhere. Storing quotes in the env vars may break other things...

We should probably fix up `activate.sh` also so it quotes the paths so will work with spaces (less likely on Unix, but still possible)...

j9ac9k: For the record, I'm no expert either.  I just stumbled across this issue when collaborating with someone that had a space in their home directory.  Thanks for the link!

I think the issue is that in the cases where there is a space somewhere in the path of `%CONDA_PREFIX%`, you're already going to have quotes in the variable, wrapping the whole expression in double quotes results in the end_quote being detected at the first quote inside the variable expression.

For your example, I'm not sure there is an issue with `%GDAL_DATA%` specifically; I thought to modify all the variable assignments that may represent paths.

I can update a comment with this PR with echo-ing out the contents of those variables if there are spaces

gillins: > I think the issue is that in the cases where there is a space somewhere in the path of %CONDA_PREFIX%, you're already going to have quotes in the variable

But there shouldn't be quotes in the variable... If there are quotes in the env var, it may work in .bat files but not if you access them from `os.environ` etc in Python (because Python won't process the quotes). I'm wondering if there is a bug in `conda` that is causing this to fail rather than `activate.bat`... I'll see if I can find some time to investigate.

gillins: I managed to get onto a Windows PC to check this. I think the only problem is with the line:
```
@if not exist %GDAL_DRIVER_PATH% (
```
It should read:
```
@if not exist "%GDAL_DRIVER_PATH%" (
```
The rest of `activate.bat` should stay the same. I think your PR was just inserting quotes that are unnecessary for anything other than working around the incorrect syntax for this line. Can you confirm this?

I note that the actual activation was working (gdal commands seemed to work for me), just the check for a `gdalplugins` directory wasn't quite right when there are spaces....

j9ac9k: I just tested the change to gdal-activate.bat you proposed and can confirm that your suggested change does make that warning go away.

I don't know if this is an issue, but checking for the presence of  %GDAL_DRIVER_PATH% doesn't seem to show (the following is console output with the change you suggested)

```
(base) C:\Users\ogi\source\repos>conda activate "C:\Users\ogi\with space"

C:\Users\ogi\source\repos>rem proj-data is installed because its license was copied over

(C:\Users\ogi\with space) C:\Users\ogi\source\repos>python
Python 3.10.5 | packaged by conda-forge | (main, Jun 14 2022, 06:57:50) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.getenv("GDAL_DRIVER_PATH")
>>> exit()

(C:\Users\ogi\with space) C:\Users\ogi\source\repos>echo %GDAL_DRIVER_PATH%
%GDAL_DRIVER_PATH%
```

If we check for the presence of some other variables...

```
(C:\Users\ogi\with space) C:\Users\ogi\source\repos>echo %GDAL_DATA%
C:\Users\ogi\with space\Library\share\gdal

(C:\Users\ogi\with space) C:\Users\ogi\source\repos>echo %CONDA_PREFIX%
C:\Users\ogi\with space
```

Perhaps this is the intended behavior that it checks the `Library\lib` directory and not `Library\bin`?

```
@set "GDAL_DRIVER_PATH=%CONDA_PREFIX%\Library\lib\gdalplugins"
```

image

gillins: You are indeed correct - the `gdalplugins` directory is under `bin` not `lib`. Do you mind fixing this up also?

I think this is new - ie gdal didn't previously create a `gdalplugins` dir on install, we just guessed that under `lib` would be a sensible place for a user to put their own drivers. If `gdal` have made a call to do it under `bin` we should follow...

j9ac9k: > You are indeed correct - the `gdalplugins` directory is under `bin` not `lib`. Do you mind fixing this up also?

Happy to do so!

j9ac9k: Follow up question, should it point to the `bin` directory on linux as well (meaning should I update those activate scripts as well?)

gillins: > Follow up question, should it point to the bin directory on linux as well

I just checked and it is under `lib` for Linux so those activate scripts are ok. If you feel like updating the `.sh` files so they also work with spaces also that would be great! I can see that in some situations (eg bash users on Windows) it would be nice to be able to cope with spaces....

j9ac9k: > I just checked and it is under `lib` for Linux so those activate scripts are ok. If you feel like updating the `.sh` files so they also work with spaces also that would be great! I can see that in some situations (eg bash users on Windows) it would be nice to be able to cope with spaces....

Least I can do 👍🏻 

Might take me a few days as I just had to give up my POSIX machine up for servicing; I'll try and test with WSL or on a Linux VM.

j9ac9k:
with regards to handling spaces in paths on linux, gdal was far from the only library in my test conda environment that had this issue.  `geotiff`,  `proj4` and `libglib` had little hick-ups.  I'll investigate submitting PRs to each of those repos.

gillins:
Thanks @j9ac9k this is a very worthwhile improvement!

j9ac9k:
> Thanks @j9ac9k this is a very worthwhile improvement!

Thanks for your patience and assistance!

zklaus:
I agree that this was important, thanks for sorting it out!

I wonder if there is something to be taken for conda-forge generally? For example, we could create the environment that runs the tests of every recipe with a space in the name. That might generate a bunch of confusing failures, though. Or we could run the tests twice, which would create perhaps undue overhead. Or we could do nothing and continue to rely on users to report problems when they appear. Thoughts?

j9ac9k:
> Or we could do nothing and continue to rely on users to report problems when they appear. Thoughts?

I won't speak to if something conda-forge should do generally; but from my perspective, I seem to have come across an edge case in my usage that does seem to be a _very_ rare case (given that nothing has been reported here).  The fix was straightforward enough that the burden on a contributor I didn't feel was particularly high (took a little effort to test fixes across platforms).

It is somewhat difficult to try and create a conda environment in a directory with a space.  I couldn't do it via `conda create -n "name with spaces"`, I couldn't do `conda create -p .` (for current directory that had a space), I had to use `conda create -p "C:\Full Path\with spaces"`.  The alternative is having `miniconda`/`miniforge`/`mambaforge` installed in a user directory, where the user name has a space in it.

zklaus:
Alright, let's just wait and see.

PR #656, opened 2022-10-31
Remove zarr

dopplershift:
Checklist

Just put a static copy of the test data in the recipe (alongside the rest) and use that. This removes a Python library from the dependency tree (where there are almost none already) in order to reduce the set of packages GDAL needs to wait on.

### DIFF
--- a/.ci_support/linux_64_openssl1.1.1.yaml
+++ b/.ci_support/linux_64_openssl1.1.1.yaml
@@ -58,7 +58,6 @@ numpy:
 - '1.21'
 - '1.20'
 - '1.20'
-- '1.20'
 openjpeg:
 - '2'
 openssl:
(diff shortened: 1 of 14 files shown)


ocefpaf:
Thanks! folks won't need to wait for https://github.com/conda-forge/numcodecs-feedstock/pull/91 with this change.

ocefpaf (approved):

zklaus:
To be clear, if I understand correctly, this does *not* remove support for the Zarr format from Gdal (which comes natively), but only removes a dependency on the `zarr` python package which was purely used for testing.

akrherz:
> To be clear, if I understand correctly, this does _not_ remove support for the Zarr format from Gdal (which comes natively), but only removes a dependency on the `zarr` python package which was purely used for testing.

That is my understanding!

ocefpaf:
Yep. Correct. That test dependency was blocking the Python 3 migration and we don't really need it, just the data for testing.

PR #678, opened 2022-11-17
Remove unused libdap4 dependency, and update pcre->pcre2

rouault:
libdap4 was necessary for the DODS drivers, but they have been removed in GDAL 3.5.0, so just drop this dependency.

pcre is deprecated. GDAL supports pcre2 too, so use the later

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -11,7 +11,7 @@ source:
     - 0001-Improve-numpy-fixing.patch
 
 build:
-  number: 2
+  number: 3
   skip_compile_pyc:
     - "share/bash-completion/completions/*.py"
 
@@ -39,7 +39,6 @@ requirements:
     - jpeg
     - json-c  # [not win]
     - kealib
-    - libdap4  # [not win]
     - libkml
     - libnetcdf
     - libpng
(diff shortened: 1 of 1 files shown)


akrherz (approved):

rouault:
> Would you have any hints?

that doesn't ring a bell at all to me. GDAL build system doesn't do anything particular with PCRE2 as far as I can see GDAL's ``cmake/modules/packages/FindPCRE2.cmake``. It just links against the pcre2 lib.
Looking at GDAL's conda-forge Mac build of https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_apis/build/builds/619034/logs/113, one can see "Found PCRE2: $PREFIX/lib/libpcre2-8.dylib", so it links against Conda-Forge's one. Nothing suspicious here.
At the worse case we could downgrade GDAL's recipe to use back pcre instead of pcre2. I did the upgrade because pcre is deprecated, but GDAL supports both.

zklaus:
Alright, let's take it at the r-gdalcubes PR for now and not change anything here.

zklaus:
Good catch, @akrherz. Seems it picked up an older build. Let's see if we can fix it.

PR #679, opened 2022-11-18
Add a libgdal-arrow-parquet package

### LINKED ISSUE #628: Add parquet-cpp dependency
Non-blocking but could be nice to have for future iterations:
- add parquet-cpp dependency (but that requires switching to CMake builds)

_Originally posted by @rouault in https://github.com/conda-forge/gdal-feedstock/issues/627#issuecomment-1126578502_

rouault:
Add a libgdal-arrow-parquet package which contains the ogr_Arrow.so/dylib/dll and ogr_Parquet.so/dylib/dll  plugins
    
The libgdal-arrow-parquet build reuses the same build directory as libgdal, hence we just to have to turn on the required dependency and drivers in the build_arrow_parquet.sh/bat scripts.
    
Fixes #628

Checklist


rouault:

akrherz:
@rouault , you are amazing! 

My two cents and I could be totally wrong here, but `arrow-cpp` is a bit of trouble in `conda-forge` these days.  For example, I am still stuck with gdal 3.4 due to dependency hell with `arrow-cpp` involved and a jam with `aws-sdk-cpp` conda-forge/arrow-cpp-feedstock#867  

So I don't know how far this PR can get until that is cleared...

akrherz:
I noticed the failed win64 openssl=3 build picked up `arrow-cpp: 0.8.0-py36_vc14_4`  yikes.

rouault:
> `arrow-cpp` involved and a jam with `aws-sdk-cpp`

Selfishly, for GDAL needs, we don't need arrow-cpp to be built against aws-sdk-cpp, as GDAL plugs its VSI I/O layer into libarrow, and thus can access AWS S3 through /vsis3/ . Would dropping aws-sdk-cpp dependency from arrow-cpp be acceptable ?

ocefpaf:
@xhochy sorry to ping you here but do you think you can help us navigate this arrow-parquet dependency for gdal?

recipe/meta.yaml
```diff
@@ -50,6 +51,7 @@ requirements:
     - libxml2
     - openjpeg
     - openssl
+    - parquet-cpp
```
xhochy, line 54: Always depend on `arrow-cpp`, `parquet-cpp` is only a meta package.

xhochy (commented):
I think it would be wise to wait until we have merged `arrow-cpp=10` as this resolves the dependency solve issue with `aws-sdk-cpp` hopefully. We can also spit the `arrow-cpp` package a bit more up to have fewer dependencies but I think the core already links to `aws-sdk-cpp`, i.e. we won't be able to get rid of that.

akrherz:

rouault:
hum, why are there attempts to build for arrow-cpp 6, 7, 8, 9 ? Shouldn't that just be against 10 ?

rouault:

akrherz:
I looked at a random build and found this:

```
"libgdal-arrow-parquet-3.6.0-h0f717e7_": {
  "recipe": {
    "arrow_cpp": "6.0.2"
```

I think we are going to be waiting for the resolution of conda-forge/conda-forge-pinning-feedstock#3793 and the openssl=3 migration ongoing and perhaps other things.

rouault:
> I looked at a random build and found this:

isn't the arrow-cpp >= 10.0 constraint I added in meta.yaml good enough ?

akrherz:
@rouault maybe the `run` entry for `arrow-cpp` needs updated too then?

rouault:
> maybe the `run` entry for `arrow-cpp` needs updated too then?

We want the libgdal-arrow-parquet package to run against the exact same (major) version of arrow-cpp as the one used for building, because of potential C++ ABI issues otherwise (this is true of other C++ dependencies, like poppler, etc). Isn't that the default behaviour ?

.ci_support/linux_64_openssl1.1.1.yaml
```diff
@@ -1,3 +1,8 @@
+arrow_cpp:
+- 6.0.2
+- 7.0.1
+- 8.0.1
+- 9.0.0
```
h-vetinari, line 5: > I looked at a random build and found this:
>
> ```
> "libgdal-arrow-parquet-3.6.0-h0f717e7_": {
>   "recipe": {
>     "arrow_cpp": "6.0.2"
> ```

That's because by default you're getting builds for 4 different arrow versions. You can add a skip to the libgdal-arrow-parquet output so that it only builds for arrow 10.x (and then wait for the [migrator](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3793) you already referenced)

recipe/meta.yaml
```diff
@@ -152,6 +154,88 @@ outputs:
...
+        - {{ compiler('c') }}
+        # libstdc++ is needed in requirements/run
+        - {{ compiler('cxx') }}
+        - pkg-config  # [not win]
+        - make  # [unix]
+        - ninja  # [win]
+        - sysroot_linux-64 2.17  # [linux64]
+      host:
+        - {{ pin_subpackage('libgdal', exact=True) }}
+        - arrow-cpp >=10.0
```
h-vetinari, line 173: Having a lower bound is fine (but should be coupled with the skip, as mentioned above), but it's better to _also_ leave an unpinned line so that future versions will be built correctly against a single arrow version.

Also, if you're fine with only building for arrow >=10, please use the new name for the library, `arrow-cpp` is just there for compatibility and will eventually be removed.

```suggestion
        - libarrow
        - libarrow >=10.0
```

recipe/meta.yaml
```diff
@@ -152,6 +154,88 @@ outputs:
...
+        - qhull
+        - sqlite
+        - tiledb
+        - xerces-c
+        - xz
+        - zlib
+        - zstd
+      run:
+        - {{ pin_subpackage('libgdal', exact=True) }}
+        - arrow-cpp
```
h-vetinari, line 212: Unless you were relying on the python bits in arrow-cpp, you don't need this anymore; libarrow has a run-export.
```suggestion
```

h-vetinari (commented):
> My two cents and I could be totally wrong here, but `arrow-cpp` is a bit of trouble in `conda-forge` these days.

This was not only in conda-forge, but due to AWS not fixing the bug that caused the pin in the first place. If it hadn't been fixed recently, I would have proposed to drop the AWS bindings due to that.

arrow 10.x now switches to 1.9.*, hope that helps.

rouault:

rouault:

.ci_support/linux_64_openssl1.1.1.yaml
rouault: I tried to add one, but conda smithy rerender wasn't happy. Probably because I'm a conda-forge noob and didn't get things right. I just went to adding libarrow only without version constraints

h-vetinari:
The pinning doesn't know about `libarrow` yet. Either you have to manually copy the migration file here (and add `use_local: true`) or wait for the migrator to be merged and the bot to open a PR here

recipe/meta.yaml
```diff
@@ -26,7 +26,6 @@ requirements:
     - {{ compiler('cxx') }}
     - sysroot_linux-64 2.17  # [linux64]
   host:
-    - arrow-cpp >=10.0
     - blosc
```
h-vetinari, line 29: For the global build requirements, you cannot switch yet, assuming you are still going to be building against previous versions of arrow.

Either you stay on arrow-cpp for everything here for a while longer (we'll keep that output working), or you do something as follows:

```suggestion
    # libarrow only exists from 10.x
    - arrow-cpp  # [arrow_cpp in ("6.0.2", "7.0.1", "8.0.1", "9.0.0")]
    - blosc
```

recipe/meta.yaml
```diff
@@ -41,6 +40,7 @@ requirements:
     - json-c  # [not win]
     - kealib
     - lerc
+    - libarrow
```
h-vetinari, line 43: ... and similarly here (if not staying with `arrow-cpp` generally)
```suggestion
    - libarrow  # [arrow_cpp == "10.0.1"]
```

h-vetinari (commented):
I cannot comment in unchanged lines, but I'd still suggest (assuming I've understood correctly) to only build the new output against arrow 10, i.e.
```diff
   - name: libgdal-arrow-parquet
     script: build_arrow_parquet.sh  # [unix]
     script: build_arrow_parquet.bat  # [win]
+    build:
+      skip: true  # [arrow_cpp != "10.0.1"]

**Comment by rouault** [comment]
@h-vetinari I must confess I'm totally confused by all those subtelties :-) Isn't the current state with https://github.com/conda-forge/gdal-feedstock/pull/679/commits/f0084745d0e0a56e60ca0c1dafe7cc5803a15ab8 good enough ? I mean the build is green...  My attempts at adding a >= 10 condition were just because I was confused to see conda-smithy rerender to create a matrix with different arrow-cpp versions. Personally, I'm fine with the current state where it seems to just use version 10.

**Comment by h-vetinari** [comment]
> @h-vetinari I must confess I'm totally confused by all those subtleties :-)

Don't worry, there's a couple tricky/unusual things here (not least the rename, sorry for the churn on that). The reason the current state is not good enough is that it will break as soon as arrow 11 is released.

Let's recap:
* for your _host_ dependencies which are present the global [pinning](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml), conda-forge will make sure your feedstock is rebuilt as newer versions of those dependencies become available.
* however, this only works if you leave the dependency unpinned (or add an extra line for that), otherwise `libfoo >= x` will _disable_ the main mechanism for the pinning, and you'll be building against whatever happens to be the newest version at the time.
* thanks to run-exports, you should then still get functional packages, but they'll be constrained to only run against the very newest library version, which makes it much harder to install those packages.
* this is particularly relevant for arrow, which - due to its importance combined, with many major (=ABI-changing) releases - is built against the four most recent versions. Your `libarrow` (due to the rename this is currently equivalent to doing `>= 10`) effectively means this feedstock will only be built against arrow 10.x, dropping 7.x - 9.x in the process.
* You'll very likely get additional breakage with the current state of this PR once https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3793 is merged, and `libarrow` actually is enforced to match the version of `arrow-cpp` (not least to catch the kind of accidental misconfiguration happening here).

**Comment by rouault** [comment]
@conda-forge-admin, please rerender

**Comment by rouault** [comment]
hum, trying latest @h-vetinari suggestions (hopefully I got those right), builds are broken. I let other maintainers follow up as this is beyond my mastering.

File: recipe/meta.yaml [hunk 7]
```diff
@@ -170,6 +174,8 @@ outputs:
         - sysroot_linux-64 2.17  # [linux64]
       host:
         - {{ pin_subpackage('libgdal', exact=True) }}
+        # libarrow only exists from 10.x
+        - arrow-cpp  # [arrow_cpp != "10.0.1"]
```
h-vetinari, line 178: Here you don't need this as you're explicitly skipping anything before arrow 10.x (I haven't caught up with the rest of the thread of why that is, but given that you seemed to be waiting for 10.x, I thought that was the intention - was it?)
```suggestion
```

h-vetinari:
> @rouault: hum, trying latest @h-vetinari suggestions (hopefully I got those right), builds are broken. I let other maintainers follow up as this is beyond my mastering.

Please see above:

> @h-vetinari: The pinning doesn't know about `libarrow` yet. Either you have to manually copy the migration file here (and add `use_local: true`) or wait for the migrator to be merged and the bot to open a PR here

h-vetinari:
I added a PR for the arrow situation in https://github.com/rouault/gdal-feedstock/pull/1.

I'm sorry I gave confusing advice before - I don't know this recipe well and didn't understand what exactly you were using arrow for. For now, I don't see a reason why you _shouldn't_ be able to support arrow versions before 10, so that's what that PR does.

Because of that, I'm sticking with the `arrow-cpp` name, because `libarrow` only exists from v10.

Key point is https://github.com/rouault/gdal-feedstock/pull/1/commits/0291c5ff9cff877e671f5a42552a72c566f2dac8 - the additional build dependencies should only be necessary for that output, which means that the extra versions don't generate separate CI jobs, but get executed sequentially in one overall job per arch.

Let me know if you have further questions. :)

recipe/meta.yaml
```diff
@@ -152,6 +153,87 @@ outputs:
       license: MIT
       license_file: LICENSE.TXT
 
+  - name: libgdal-arrow-parquet
+    script: build_arrow_parquet.sh  # [unix]
+    script: build_arrow_parquet.bat  # [win]
```
h-vetinari, line 158:
```suggestion
    script: build_arrow_parquet.bat  # [win]
    build:
      # currently, openssl 3 only works for arrow>=10; this may change if certain changes are
      # backported; tracked in e.g. https://github.com/conda-forge/arrow-cpp-feedstock/pull/874
      skip: true  # [(openssl == "3")  and (arrow_cpp != "10.0.1")]
```

h-vetinari (commented):
ah, damn, forgot that openssl 3 is only compatible with arrow >=10 (so far; will depend if we backport the aws-sdk-cpp bump to the previous branches, which depends on no-one reporting regressions on aws with v10)

In the meantime, I'd suggest to add a skip as below

rouault:

h-vetinari:
I don't understand why the skip is not taking effect. conda is still trying to solve:
```
INFO:conda_build.metadata:Attempting to finalize metadata for libgdal-arrow-parquet
Attempting to finalize metadata for libgdal-arrow-parquet
Mamba failed to solve:
 - [...]
 - arrow-cpp 7.0.1.*
 - [...]
 - openssl 3.*
```

rouault:

rouault:

rouault:

recipe/meta.yaml
```diff
@@ -152,6 +153,91 @@ outputs:
...
+    build:
+      # currently, openssl 3 only works for arrow>=10; this may change if certain changes are
+      # backported; tracked in e.g. https://github.com/conda-forge/arrow-cpp-feedstock/pull/874
+      skip: true  # [(openssl == "3")  and (arrow_cpp != "10.0.1")]
```
h-vetinari, line 162: I think we could skip openssl 3 for now...

```suggestion
      # skip openssl 3 pending https://github.com/conda-forge/arrow-cpp-feedstock/issues/918;
      # alternatively only dependon arrow > 10 (conda seems to get confused with a combined skip)
      skip: true  # [openssl == "3"]
```

rouault:

jorisvandenbossche:
(if the older arrow versions give problems, I think it's also perfectly fine to only have this for arrow-cpp 10 for now, since it's a new install feature for conda users anyway)

h-vetinari:
@rouault, your last few commits make no sense, as they will be immediately undone whenever the feedstock gets rerendered. Also, what's the issue with curl? If you don't want it, you'll have to remove it from the host-dependencies.

By the argument of Joris, I think it would be much easier to just skip the new output for older arrow version

recipe/meta.yaml
```diff
@@ -152,6 +153,91 @@ outputs:
       license: MIT
       license_file: LICENSE.TXT
 
+  - name: libgdal-arrow-parquet
+    script: build_arrow_parquet.sh  # [unix]
+    script: build_arrow_parquet.bat  # [win]
+    build:
+      # skip openssl 3 pending https://github.com/conda-forge/arrow-cpp-feedstock/issues/918;
+      # alternatively only dependon arrow > 10 (conda seems to get confused with a combined skip)
+      skip: true  # [openssl == "3"]
```
h-vetinari, line 162:
```suggestion
      # only for libarrow > 10
      skip: true  # [arrow_cpp != "10.0.1"]
```

rouault:
> your last few commits make no sense, as they will be immediately undone whenever the feedstock gets rerendered. Also, what's the issue with curl?

Probably. I told you I was a conda-forge noob ;-) well, I did them based on the error messages in failed builds, and looking at the diff of this PR against main. The added curl: 7 constraint was new to this PR and totally unrelated. We do want curl, but some package didn't like that constraint.

rouault:
Hooray: green build!

akrherz (approved):

PR #685, opened 2022-11-20
 Do not rebuild libgdal when creating python packages (fixes #662) 

### LINKED ISSUE #662: use cmake python_generated_files target to save build time
> Would it make sense to do the SWIG_REGENERATE thing instead of the `cmake --build`?

No, SWIG_REGENERATE_PYTHON option will not help if you don't run the python_wheel target. 
But I missed that python_wheel does build libgdal as a dependency.
OK, I'm proposing https://github.com/OSGeo/gdal/pull/6704 which adds a python_generated_files target that should do what you want, and doesn't require building libgdal. But you will need to provide to setup.py the path to the GDAL include and library, because by default it will point to in-tree paths/files that don't exist if you just run this python_generated_files target.

_Originally posted by @rouault in https://github.com/conda-forge/gdal-feedstock/pull/651#discussion_r1022896222_

rouault:
and add qhull dependency instead of depending on internal qhull copy

Checklist


gillins:
Thanks @rouault ! Do you want to add yourself as a maintainer?

gillins (approved):

rouault:
> Do you want to add yourself as a maintainer?

no, I'm already declared maintainer of too many things :-)

ocefpaf:
> > Do you want to add yourself as a maintainer?
> 
> no, I'm already declared maintainer of too many things :-)

While the last thing I want is to give you more things to worry about. Having you as a maintainer here is also a way to credit all the work your are putting into the conda-forge package lately. There are no real duties for a maintainer, the goal is to increase the bus factor so others can pick up the work when someone cannot. The main downside are the @/mentions you would get due to the use of CODEOWNERS in our feedstocks.

PR #703, opened 2022-12-15
Update to GDAL 3.6.1

rouault:
Checklist
* [X] Reset the build number to `0` (if the version changed)


rouault:

ocefpaf (approved):

rouault:
The GDAL project has officially retracted 3.6.0 due to the issue outlined at top of https://github.com/OSGeo/gdal/blob/v3.6.1/NEWS.md . Would blacklisting Conda-Forge GDAL 3.6.0 packages would be too extreme... ?

ocefpaf:
> The GDAL project has officially retracted 3.6.0 due to the issue outlined at top of https://github.com/OSGeo/gdal/blob/v3.6.1/NEWS.md . Would blacklisting Conda-Forge GDAL 3.6.0 packages would be too extreme... ?

Marking them as broken is probably the right thing to do in this case. However, it is not an easy decision and we need to weight the pros and cons:

Cons:
- we have 14 variants of gdal 3.6.0 out there, marking them all as broken means that we won't cover all the past dependency migrations that happened to 3.6.0, only the "last set."
- Only GeoPackage is affect, right, so most of gdal's 3.6.0 functionality are still usable.

Pros:
- Folks using latest version, or locked environments, on get hit by this b/c 3.6.1 seems to be 100% ABI compatible with 3.6.0 (please correct me if I'm wrong here).
- GeoPackage is a pretty important feature in gdal, having it broken may outweigh keeping 3.6.0.

zklaus:
Another alternative might be to implement the work-around from the release notes in the activation script of a new build of 3.6.0.

ocefpaf:
> Another alternative might be to implement the work-around from the release notes in the activation script of a new build of 3.6.0.

Yep. Warning the users is fine with me. We can try conda's new messaging mechanism.

rouault:
> Only GeoPackage is affect, 

yes and for big enough layers (typically 100 000 features or more, although I'm not completely sure of the minimum threshold which might depend on how GDAL API is actually used)

> * seems to be 100% ABI compatible with 3.6.0

it should! (always hard to formally acertain this). But 3.6.1 has ~ 60 other bugfixes, some of them resulting in intended changes in output (output of gdalwarp slightly changes in 3.6.1 on some border pixels).
I'm not sure if that might help (probably not as that wouldn't fix released packages), but fixing the GeoPackage breakage against 3.6.0 is a 2-liner patch: https://github.com/OSGeo/gdal/commit/3f5f6225fe82e0c2e0241e4f66bfb861cdf4fe9d

ocefpaf:
> Another alternative might be to implement the work-around from the release notes in the activation script of a new build of 3.6.0.

Now that I stopped to think about it we cannot re-do the past packages. We don't have a messaging mechanism that patches the existing packages.

> I'm not sure if that might help (probably not as that wouldn't fix released packages), but fixing the GeoPackage breakage against 3.6.0 is a 2-liner patch: https://github.com/OSGeo/gdal/commit/3f5f6225fe82e0c2e0241e4f66bfb861cdf4fe9d

Patching 3.6.0 with that fix is, IMO, not worth it. Better to just update to 3.6.1.

> yes and for big enough layers (typically 100 000 features or more, although I'm not completely sure of the minimum threshold which might depend on how GDAL API is actually used)

In light of this maybe we can just add an entry in our news feed and leave the packages as-is.

PR #788, opened 2023-07-19
[3.6.x] Fix deactivate.sh

xylar:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -11,7 +11,7 @@ source:
     - flatgeobuffcxx17.patch
 
 build:
-  number: 9
+  number: 10
   skip_compile_pyc:
     - share/bash-completion/completions/*.py
 
--- a/recipe/scripts/deactivate.sh
+++ b/recipe/scripts/deactivate.sh
@@ -8,7 +8,7 @@ if [ -n "${_CONDA_SET_GDAL_DATA:-}" ]; then
 fi
 
 unset GDAL_DRIVER_PATH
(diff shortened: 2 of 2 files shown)


xylar:

xylar:
Same here, feel free to merge once CI finishes unless there's another typo I missed.

rouault:
Aren't those scripts tested by CI ?
Perhaps a ``set -eu`` would be needed in them ?

xylar:
@rouault, I'm not sure about a `set -eu`.  That's certainly not a standard practice on conda-forge and I thought I had the sense that was a bad idea when you source a script as opposed to run it.

> Aren't those scripts tested by CI ?

The activate scripts are certainly tested.  The deactivate scripts maybe not or any errors they produce may simply be ignored, not sure about that.  I guess that's where your `set -eu` would come in.  I'll leave it to folks with more shell knowledge to decide if that can be added.

xylar:
So looking into this further, I do think `set -eu` in a sourced script will cause the shell to exit completely when there's an error, see https://stackoverflow.com/a/24460931/7728169.  So it doesn't seem like that's something we can add.

Anyone have any other input on this?  Otherwise, I'd like to merge soon because this bug is likely to break deactivation for folks using the latest gdal.

rouault:
> Anyone have any other input on this?

Please merge the fix. And thanks for your valid points on why set -eu wouldn't be appropriate

xylar:
Thanks @rouault

Issue #834, opened 2023-11-07
No tests for .py GDAL utilities?


plimkilde:
### Comment:

It seems there are no tests in this feedstock for the GDAL apps implemented in Python (`gdal_calc.py`, `gdal_fillnodata.py` etc.). On Windows in particular, implementation of Python commands can be finicky - currently, calling these utilities relies on .py file association, which can be messy with multiple environments. In contrast, `pip` installation of an entry point creates a small .exe file in the environment, which does not cause issues if .py files are associated with an editor or similar.

What would be needed to add tests for the Python utilities - simply adding `gdal_calc.py --help` etc. to `run_test.bat`/`run_test.sh`?

And would it be acceptable to use a `pip`-based install instead?

gillins:
Indeed finicky on Windows! Regarding entry points, I haven't had much luck creating entry points some something with an extension already (ie `gdal_calc.py`, creating a `gdal_calc.py.exe` doesn't work - Windows doesn't search for a `.exe` if already has an extension last time I checked). Dropping the `.py` means we'll be out of step with the documentation (and Unix). 
But yes, we should test that these run in `run_test.bat`/`run_test.sh` - a PR would be most welcome.

@rouault are there any plans to move to an entry point install for these scripts?

rouault:
> @rouault are there any plans to move to an entry point install for these scripts?

hum Python setup.py magic isn't totally an area I'm familiar with. May I ask @idanmiara if he sees some counter-indication in adding a entry_points keyword in https://github.com/OSGeo/gdal/blob/master/swig/python/setup.py.in while there is one in https://github.com/OSGeo/gdal/blob/master/swig/python/gdal-utils/setup.py ? I'm afaid that the GDAL Python bindings and the gdal-utils package would then collide regarding the wrapper scripts... ?

gillins:
Thanks @rouault, isn't there quite a bit of collision between GDAL Python and gdal-utils anyway? Happy to defer to you and @idanmiara but moving to entry points for GDAL Python seems like a good idea (while keeping the old scripts around for a bit). This would definitely fix the unusual behaviour on Windows. I can do a PR if you like.

idanmiara:
Back then I've created the [batch wrapper creator](https://github.com/OSGeo/gdal/blob/master/swig/python/gdal-utils/osgeo_utils/auxiliary/batch_creator.py) which was mostly superseded by `entry_points` and `console_scripts` by @maphew. see:
[...]

PR #835, opened 2023-11-07
Add tests for .py utilities

plimkilde:
Checklist

This adds tests for the Python-based GDAL utilities (`gdal_calc.py`, `gdal_fillnodata.py` etc.), see #834.

Interestingly, local testing revealed spotty support for both `--help` and `--version`.


xylar:
@plimkilde, `run_test.sh` and `run_test.bat` get run when building `libgdal`, not `gdal` so the python tools aren't available.  I think you want to add these tests directly in `meta.yaml` here:
https://github.com/plimkilde/gdal-feedstock/blob/py-utility-tests/recipe/meta.yaml#L285-L286
```
      commands:
        - python run_test.py
        - rgb2pct.py --help
        - pct2rgb.py --help
        - gdalattachpct.py --help
...
```
This should work for all architectures.

recipe/meta.yaml
```diff
@@ -284,6 +284,27 @@ outputs:
         - osgeo._osr
       commands:
         - python run_test.py
+        # Check .py GDAL utilities
+        # --help not implemented as documented for gdal_merge.py, gdal_retile.py, gdal_sieve.py, gdalmove.py, gdal_edit.py, gdal_pansharpen.py, gdalcompare.py, ogrmerge.py?
+        # also, --version does not work with e.g. rgb2pct.py...
+        - rgb2pct.py --help
+        - pct2rgb.py --help
+        - gdalattachpct.py --help
```
xylar, line 292: Getting an error on this one:
```
gdalattachpct.py --help
Usage: gdalattachpct.py <pctfile> <infile> <outfile>
```
We might have to change this to:
```suggestion
        - if not exist %PREFIX%\\Library\\bin\\gdalattachpct.py exit 1  # [win]
        - test -f $PREFIX/bin/gdalattachpct.py  # [unix]
```
though I'm certainly up for smarter suggestions if anyone has one.

plimkilde: Done (it wasn't entirely clear to me in local testing that it was an error). The better suggestion, I guess, would be to properly fix `--help` in upstream GDAL for this and other scripts.

xylar: It would be even better to make them entry points instead of scripts so they work as expected in Windows as discussed in #834

recipe/meta.yaml
```diff
@@ -284,6 +284,29 @@ outputs:
...
+        # also, --version does not work with e.g. rgb2pct.py...
+        - rgb2pct.py --help
+        - pct2rgb.py --help
+        # Check for existence of gdalattachpct.py as neither --help nor --version seem to work
+        - if not exist %PREFIX%\\Library\\bin\\gdalattachpct.py exit 1  # [win]
+        - test -f $PREFIX/bin/gdalattachpct.py  # [unix]
+        - gdal_merge.py --version
+        - gdal2tiles.py --help
+        - gdal2xyz.py --help
+        - gdal_retile.py --version
```
xylar, line 298: Oh, dear.  Seeing:
```
gdal_retile.py --version
GDAL 3.7.3, released 2023/10/30
```
but then it seems like it's exiting with an error code.

plimkilde: Yep 😩

rouault: will be addressed per https://github.com/OSGeo/gdal/pull/8676

xylar:
Aaargh!  Windows testing is just saying:
```
Tests failed for gdal-3.7.3-py312hb5f3cee_3.conda - moving package to D:\bld\broken
```
but not telling us why.  Anyone from @conda-forge/gdal have an suggestions for how to get more helpful output here?  I really don't speak Windows.

plimkilde:
> Aaargh! Windows testing is just saying:
> 
> ```
> Tests failed for gdal-3.7.3-py312hb5f3cee_3.conda - moving package to D:\bld\broken
> ```
> 
> but not telling us why. Anyone from @conda-forge/gdal have an suggestions for how to get more helpful output here? I really don't speak Windows.

Hm, I don't see any of the Windows builds successfully running to completion. I'm wondering if we are checking for existence in the correct directory?

xylar:
> I'm wondering if we are checking for existence in the correct directory?

I believe so.

recipe/meta.yaml
```diff
@@ -284,6 +284,42 @@ outputs:
...
       commands:
         - python run_test.py
+        # Check .py GDAL utilities
+        # --help not implemented as documented for gdalattachpct.py, gdal_merge.py, gdal_retile.py, gdal_sieve.py, gdalmove.py, gdal_edit.py, gdal_pansharpen.py, gdalcompare.py, ogrmerge.py?
+        # also, --version does not work with e.g. rgb2pct.py...
+        # In addition, some scripts are checked only for existence (gdalattachpct.py, gdal_retile.py, gdal_proximity.py) as they unexpectedly return error codes.
+        - rgb2pct.py --help
+        - pct2rgb.py --help
+        # Neither --help nor --version seem to work
+        # - if not exist %PREFIX%\\Library\\bin\\gdalattachpct.py exit 1  # [win]
```
xylar, line 294: Maybe try this next:
```suggestion
        # - if not exist %LIBRARY_BIN%\\gdalattachpct.py exit 1  # [win]
```

xylar: But I believe that's the same.

zklaus:
If we need more info about how things are failing, we could put the tests in their own scripts, say `test_python_scripts.{sh,bat}` and add `set -xe` and `@echo on`, respectively.

xylar:
@zklaus, that's a great suggestion.

gillins:
I think on Windows, you'll need to run:
```
%PYTHON% %LIBRARY_BIN%\rgb2pct.py --help  # [win]
```
etc. It's probably popped up a dialog box wanting to know how to execute a `.py` file and that's why the build if failing (?). Windows doesn't have the shebang thing so needs to be told how to execute the file....

plimkilde:
> I think on Windows, you'll need to run:
> 
> ```
> %PYTHON% %LIBRARY_BIN%\rgb2pct.py --help  # [win]
> ```
> 
> etc. It's probably popped up a dialog box wanting to know how to execute a `.py` file and that's why the build if failing (?). Windows doesn't have the shebang thing so needs to be told how to execute the file....

Trouble is, then we're not testing the documented commands 😕

gillins:
> Trouble is, then we're not testing the documented commands 😕

What's documented has never worked on Windows as-is. This has been a source of much confusion when I've been helping Windows users. I've always had to tell people to prefix everything with `python` and then hardcode the path to the script (which then breaks when they change envs).
Entry points would be the way forward for cross platform (conda can generate these if asked: https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#python-entry-points). But then we'd have to lose the extension, which again, wouldn't match the docs. Argghh.

plimkilde:
> > Trouble is, then we're not testing the documented commands 😕
> 
> What's documented has never worked on Windows as-is. This has been a source of much confusion when I've been helping Windows users. I've always had to tell people to prefix everything with `python` and then hardcode the path to the script (which then breaks when they change envs). Entry points would be the way forward for cross platform (conda can generate these if asked: https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#python-entry-points). But then we'd have to lose the extension, which again, wouldn't match the docs. Argghh.

I don't feel inclined to write a passing test for something that doesn't work, so maybe we should simply limit the scope of this PR to simply adding tests of what works (i.e. Unix), then enabling Windows testing in a later PR when a solution is ready.

plimkilde:
Hm, it seems to pass on Windows with existence checks disabled. As suggested above, I've tried to move them into a separate .bat file to have more details.

recipe/meta.yaml
```diff
@@ -284,6 +285,43 @@ outputs:
...
+        # - if not exist %PREFIX%\\Library\\bin\\gdal_pansharpen.py exit 1  # [win]
+        - test -f $PREFIX/bin/gdal_pansharpen.py  # [unix]
+        - gdalcompare.py --version
+        # Errors out on --version despite printing GDAL version
+        # - if not exist %PREFIX%\\Library\\bin\\ogrmerge.py exit 1  # [win]
+        - test -f $PREFIX/bin/ogrmerge.py  # [unix]        
+        # Returns error despite printing help
+        # - if not exist %PREFIX%\\Library\\bin\\ogr_layer_algebra.py exit 1  # [win]
+        - test -f $PREFIX/bin/ogr_layer_algebra.py  # [unix]
+        - test_script_exist.bat # [win]
```
xylar, line 324:
```suggestion
        - test_script_exist.bat  # [win]
```

recipe/test_script_exist.bat
```diff
@@ -0,0 +1,25 @@
+@echo on
+
+REM As a workaround for GDAL apps unexpectedly returning error codes, this
+REM script checks only for the existence of the .py files below.
+
+echo Test existence of gdalattachpct.py...
+if not exist %PREFIX%\\Library\\bin\\gdalattachpct.py exit 1
```
xylar, line 7: We could tidy up with `%LIBRARY_BIN%` but let's see if this works first.

plimkilde: That does not seem to make any difference

xylar: As I said, it's just tidier.  If the file isn't one place, it isn't the other because they point the same place.  If the test is failing, the file presumably really isn't installed there.

recipe/test_script_exist.bat
```diff
@@ -0,0 +1,25 @@
+REM As a workaround for GDAL apps unexpectedly returning error codes, this
+REM script checks only for the existence of the .py files below.
+
+@echo on
+
+echo Test existence of gdalattachpct.py...
+if not exist %LIBRARY_BIN%\\gdalattachpct.py exit 1
```
gillins, line 7: Single back slash should do it. You just need double in the yaml.

plimkilde: True, though it should work in practice in most cases.

recipe/test_script_exist.bat
```diff
@@ -0,0 +1,21 @@
+@echo off
+REM As a workaround for GDAL apps unexpectedly returning error codes, this
```
rouault, line 2: Note: GDAL 3.8.0RC2 has been issued with the fix for that issue

xylar: @rouault, is there also a plan to move from scripts to entry points to better support Windows?

xylar (approved):
@plimkilde, thanks for persisting through this one!  Nice work!

recipe/test_script_exist.bat
rouault: @xylar I'm waiting for feedback. See https://github.com/conda-forge/gdal-feedstock/issues/834#issuecomment-1798334805

xylar: Ah, sorry, I had noticed that comment and just forgot.  Thanks!

gillins:
Thanks @plimkilde !

PR #887, opened 2024-02-03
Switch to openssl 3.0 in host

### LINKED ISSUE #886: openssl reference in GDAL recipe?
In 2021, #543 had us add an `openssl` to the `host` section. It can be seen at https://github.com/conda-forge/gdal-feedstock/blob/main/recipe/meta.yaml#L274C11-L274C66

I have a situation where I want to run Conda Forge stuff on Google Colab, which has OpenSSL 3.0. I'd like to be able to install that, but it seems the GDAL package has a pin:

```
mamba install pdal=2.6.2 openssl=3.0
```
> The following packages are incompatible
├─ openssl 3.0**  is requested and can be installed;
└─ pdal 2.6.2**  is not installable because it requires
   ├─ libgdal >=3.8.3,<3.9.0a0 , which requires
   │  └─ openssl >=3.2.0,<4.0a0 , which conflicts with any installable versions previously reported;
   └─ openssl >=3.2.0,<4.0a0 , which conflicts with any installable versions previously reported

Do we really still need `openssl` in the `host` section?

xylar:
Drop openssl from gdal (but not libgdal)

Checklist

closes #886

### DIFF
--- a/.ci_support/linux_64_.yaml
+++ b/.ci_support/linux_64_.yaml
@@ -80,7 +80,7 @@ numpy:
 openjpeg:
 - '2'
 openssl:
-- '3'
+- '3.0'
 pcre2:
 - '10.42'
 pin_run_as_build:
--- a/.ci_support/linux_aarch64_.yaml
+++ b/.ci_support/linux_aarch64_.yaml
@@ -84,7 +84,7 @@ numpy:
 openjpeg:
 - '2'
 openssl:
-- '3'
+- '3.0'
 pcre2:
 - '10.42'
 pin_run_as_build:
--- a/.ci_support/linux_ppc64le_.yaml
+++ b/.ci_support/linux_ppc64le_.yaml
@@ -80,7 +80,7 @@ numpy:
 openjpeg:
 - '2'
 openssl:
-- '3'
+- '3.0'
 pcre2:
 - '10.42'
 pin_run_as_build:
--- a/.ci_support/osx_64_.yaml
+++ b/.ci_support/osx_64_.yaml
@@ -78,7 +78,7 @@ numpy:
 openjpeg:
 - '2'
 openssl:
-- '3'
+- '3.0'
 pcre2:
 - '10.42'
 pin_run_as_build:
--- a/.ci_support/osx_arm64_.yaml
+++ b/.ci_support/osx_arm64_.yaml
@@ -78,7 +78,7 @@ numpy:
 openjpeg:
 - '2'
 openssl:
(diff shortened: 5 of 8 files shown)


xylar:

akrherz:
I don't think this will help much as there's till plenty of downstream requirements of libgdal that will have `openssl >=3.2` requirements, ie postgresql ...

I also don't understand where within the build system the `openssl>=3.2` gets set given what is in [conda-forge-pinning](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L667) and what [openssl run_exports](https://github.com/conda-forge/openssl-feedstock/blob/main/recipe/meta.yaml#L33) are set to.

xylar:
@akrherz, the downstream dependencies aren't the problem.  It's the upstream dependencies that matter and, yes, there are too many of them to let us go back to an earlier openssl :-(

xylar:
> I also don't understand where within the build system the openssl>=3.2 gets set given what is in [conda-forge-pinning](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L667) and what [openssl run_exports](https://github.com/conda-forge/openssl-feedstock/blob/main/recipe/meta.yaml#L33) are set to.

I'm pretty sure that's a fundamental property of `conda-build`.  It pins the run version to be `>=` the version used to build the package.

xylar:
I'm going to close this, since it's not a practical solution.

rouault:
> It pins the run version to be `>=` the version used to build the package.

IMHO, that's a very reasonable feature, not a bug.  For example PROJ 9.2 is backwards compatible of PROJ 9.1, ie a GDAL built against PROJ 9.1 will run against PROJ 9.2, but if you build GDAL against PROJ 9.2, it may use new symbols introduced in PROJ 9.2 and not present in 9.1.  For openssl, we don't have code that tests openssl >= 3, but perhaps there are #define macros that resolve to different function names, or tricks like that, which also make it impractical to run a GDAL built against openssl 3.2 with 3.0

xylar:
@rouault, I completely agree.  It's based on the assumption that packages are forward compatible but not necessarily backward compatible, which is a common feature of minor revisions.

I don't think there would be a problem with going back to openssl 3.0 at build time *here*.  We would still be compatible with v3.2 at runtime.  But we can't do that without also doing it for all the upstream dependencies that currently use openssl > 3.0 and that's not a project I'm willing to take on.

PR #934, opened 2024-05-13
Test entry points, call all Python apps with --help

plimkilde:
Checklist
* [X] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [X] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

Closes #834 

Now that GDAL 3.9 has landed, we can test the proper entry points (Python apps, called without the .py extension). Also incorporated are changes (as a result of upstream GDAL fixes) so that all of these apps are now tested with `--help`, see https://github.com/OSGeo/gdal/issues/8717.

There has been a bit of churn on GDAL upstream regarding these entry points, see:
- https://github.com/OSGeo/gdal/pull/8718
- https://github.com/OSGeo/gdal/pull/8774
- https://github.com/OSGeo/gdal/issues/8811
- https://github.com/OSGeo/gdal/pull/8815
- https://github.com/OSGeo/gdal/pull/8824

The commands including `.py` are still the ones shown in GDAL docs, so for now I test both e.g. `gdal_calc.py` and `gdal_calc`.

PR #941, opened 2024-05-29
Remove tests for programs with .py extension

plimkilde:
Checklist

Just noticed that the `.py` extension has recently been removed from the [GDAL docs](https://gdal.org/programs/index.html) of Python programs. As a followup to #934, ditch the tests of calls with `.py`.

There are currently 3 programs for which the docs still have `.py`, I've opened [this PR](https://github.com/OSGeo/gdal/pull/10053) to fix that.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -9,7 +9,7 @@ source:
   sha256: 577f80e9d14ff7c90b6bfbc34201652b4546700c01543efb4f4c3050e0b3fda2
 
 build:
-  number: 5
+  number: 6
   skip_compile_pyc:
     - share/bash-completion/completions/*.py
   skip: true  # [ppc64le]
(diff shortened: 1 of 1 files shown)


gillins:
So this would break code for anyone still calling the `.py` extension versions? Perhaps we should wait a little longer before doing this. We perhaps need to print a warning when the old ones are called, not sure how easy that would be....

rouault:
> So this would break code for anyone still calling the `.py` extension versions?

no, the .py scripts are still there, and there's no plan to remove them. It is just that the recommended way is to use the non-suffixed launcher.

gillins:
Thanks @rouault - ah right it's just the tests that we are removing. But - perhaps they need to stay in until they are removed from GDAL? So we can be sure they aren't broken?

rouault:
> But - perhaps they need to stay in until they are removed from GDAL?

I'm fine with either option.

Issue #722, opened 2023-01-23
Creating a lightweight version of gdal dependencies


RenaudLN:
### Comment:

Currently the `libgdal` feedstock includes a large number of dependencies that are not strictly required to use GDAL (e.g. poppler, postgres, ...). This bloats image sizes everytime we want to install a GDAL-related package like `geopandas`.

Would it be possible to create a lightweight version of the feedstock?

Maybe something like

```yaml
  - name: libgdal-lite
    script: install_lib.sh  # [unix], probably needs to be updated?
    script: install_lib.bat  # [win], probably needs to be updated?
    build:
      run_exports:
        # no idea, going with minor pin
        - {{ pin_subpackage('libgdal', max_pin='x.x') }}
    requirements:
      build:
        - cmake
        # ranlib used at install time
        - {{ compiler('c') }}
        # libstdc++ is needed in requirements/run
        - {{ compiler('cxx') }}
        - pkg-config  # [not win]
        - make  # [unix]
        - ninja  # [win]
        - sysroot_linux-64 2.17  # [linux64]
      host:
        - blosc
        - geos
        - icu
        - json-c  # [not win]
        - libcurl
        - libdeflate
        - libiconv
        - libkml
        - libnetcdf
        - libpng
        - libpq
        - libspatialite
        - libtiff
        - libuuid  # [linux]
        - libwebp-base
        - libxml2
        - lz4-c
        - proj
        - zlib
        - zstd
      run:
        - json-c  # [not win]
        - libpq
        - libspatialite
        - libuuid  # [linux]
        - libwebp-base
        - proj
        - zstd
    test:
      files:
        - test_data
        - run_test.bat  # probably needs to be updated?
        - run_test.sh  # probably needs to be updated?
    about:
      summary: The Geospatial Data Abstraction Library (GDAL)
      license: MIT
      license_file: LICENSE.TXT
```

ocefpaf:
I considered this is the past. I'm not sure what are the implications of mixing the variants and/or making them incompatible with one another. For example, if we build fiona with the lite version but install the full, will it work as expected? Maybe if all the symbols are there but someone looking for a codec in the full version may be frustrated? Or will it work? I don't know the answer to all these questions.

One could create lite versions of all the packages downstream but that would be quite confusing to the users.

Maybe we can reduce the size of the end package with some binary stripping and removing/moving some files that are not usually used, like docs, etc.

rouault:
[...]

PR #975, opened 2024-08-13
Fix libgdal-pdf that fails to work

rouault:
due to poppler not being found when building libgdal-core

which causes the plugin to emit the following and not being functional:
```
Warning 1: Driver PDF declares DMD_OPENOPTIONLIST whereas its proxy doesn't declare it or with a different value
Warning 1: Driver PDF declares DCAP_OPEN whereas its proxy doesn't declare it or with a different value
Warning 1: Driver PDF declares DMD_SUBDATASETS whereas its proxy doesn't declare it or with a different value
Warning 1: Proxy driver PDF declares a pfnIdentify callback whereas the real driver does not.
Warning 1: Driver PDF declares a pfnOpen callback whereas its proxy doest not declare DCAP_OPEN
```


akrherz:
poppler was one of the main things that drove #722.  Does poppler just need to be present at build time and not at run?

akrherz:
> just at build time.

I believe adding it as a dep will cause it to be added as a run dep, unless we do some tricks.  Will research this more soon and hopefully I am wrong :)

rouault:
> Technically it doesn't even need to be there. Just having -DHAVE_POPPLER in CXXFLAGS of libgdal-core should be enough

trying that approach in https://github.com/conda-forge/gdal-feedstock/pull/976

xylar:
> I believe adding it as a dep will cause it to be added as a run dep, unless we do some tricks. 

I don't think so.  I didn't see any `run_exports`:
https://github.com/conda-forge/poppler-feedstock/blob/main/recipe/meta.yaml

akrherz:
Can a simple test be added to the `libgdal-pdf` artifact section to ensure that it is working?

rouault:
> Can a simple test be added to the `libgdal-pdf` artifact section to ensure that it is working?

attempted in #976 with 41f671383fa4ab65fdf7bc4fb66faaf0bf6ede0b

rouault:
closing as superseded per #976

PR #1017, opened 2024-12-04
meta.yml: pin libgdal-xxxx plugins to same x.y.z version of libgdal-core

### LINKED ISSUE #1016: Are we pinning GDAL plugins correctly to libgdal-core?
### Comment:

I am likely just driving myself crazy with understanding how pinning works, but my present doubt is how we pin plugins.  For example

libgdal-pdf
```
libgdal-core >=3.10
```

For a current example, that implies that one could install `libgdal-pdf-3.9.0` and `libgdal-core-3.10.0`:

```
$ mamba create -n dev python=3.13 libgdal-core=3.10 libgdal-pdf=3.9
  + libgdal-core                     3.10.0  h7250d82_3          conda-forge       11MB
  + libgdal-pdf                       3.9.3  hb578332_5          conda-forge      669kB

$ lib/gdalplugins/gdal_PDF.so is linked to libgdal.so.35, which does not exist.
```

Anybody with thoughts on this?

rouault:
Fixes #1016

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -11,7 +11,7 @@ source:
     - 000_cmake.patch  # [osx]
 
 build:
-  number: 4
+  number: 5
   skip_compile_pyc:
     - share/bash-completion/completions/*.py
   ignore_run_exports_from:
@@ -112,7 +112,7 @@ outputs:
         - libarrow-dataset  # [libarrow != 13]
         - libparquet        # [libarrow != 13]
       run:
-        - libgdal-core >={{ ".".join(version.split(".")[:2]) }}
+        - {{ pin_subpackage('libgdal-core', exact=True) }}
     test:
       commands:
         - test -f ${PREFIX}/lib/gdalplugins/ogr_Arrow${SHLIB_EXT}        # [unix]
@@ -154,7 +154,7 @@ outputs:
         - openjpeg
         - expat
       run:
-        - libgdal-core >={{ ".".join(version.split(".")[:2]) }}
+        - {{ pin_subpackage('libgdal-core', exact=True) }}
     test:
       commands:
         - test -f ${PREFIX}/lib/gdalplugins/${GDAL_PLUGIN_TYPE}_${GDAL_PLUGIN_NAME}${SHLIB_EXT}      # [unix]
@@ -194,7 +194,7 @@ outputs:
         - expat
         - poppler
       run:
-        - libgdal-core >={{ ".".join(version.split(".")[:2]) }}
+        - {{ pin_subpackage('libgdal-core', exact=True) }}
     test:
       files:
         - test_data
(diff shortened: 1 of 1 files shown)


rouault:

recipe/meta.yaml
```diff
@@ -112,7 +112,7 @@ outputs:
         - libarrow-dataset  # [libarrow != 13]
         - libparquet        # [libarrow != 13]
       run:
-        - libgdal-core >={{ ".".join(version.split(".")[:2]) }}
+        - libgdal-core =={{ ".".join(version.split(".")[:3]) }}
```
xylar, line 115: I believe the preferred way to to this would be:
```suggestion
        - {{ pin_subpackage('libgdal-core', exact=True) }}
```

xylar:
I think this looks great other than the minor clean-up I suggested above.

I just want to make sure I understand the desired behavior.  We have run_exports on `libgdal-core` set to `x.x`:
https://github.com/conda-forge/gdal-feedstock/blob/9a11d7c05793a4fd5cca96705ac62be08b113005/recipe/meta.yaml#L24-L25
So it the expectation is that newer patch versions will still work at runtime for a package that was built with an older patch version.

@rouault, it seems like you are suggesting that, nonetheless, we want all the split packages to come from the exact same patch version because it would not necessarily be safe to have libgdal subpackages from different patch versions installed at runtime.  That makes sense to me.  If you update one, you have to update them all.

rouault:
I've just amended the commit to use ``{{ pin_subpackage('libgdal-core', exact=True) }}`` instead of ``libgdal-core =={{ ".".join(version.split(".")[:3]) }}``

should I also change ``{{ pin_subpackage('libgdal-core', max_pin='x.x') }}`` at line 25? I don't know what is the scope of this line.

> it seems like you are suggesting that, nonetheless, we want all the split packages to come from the exact same patch version because it would not necessarily be safe to have libgdal subpackages from different patch versions installed at runtime. That makes sense to me.

in theory a plugin built against 3.10.0 should be able to run against 3.10.1, but that's not really tested upstream, and could result in hard-to-debug situations. I believe it is clearer/safer that if someone updates their libgdal-core, they update their plugins to the same version.

xylar:
> should I also change {{ pin_subpackage('libgdal-core', max_pin='x.x') }} at line 25? I don't know what is the scope of this line.

@rouault, no I think that should stay unless we have a good reason to change it.  If we change this, it means we have to do a (lib)gdal migration on conda-forge each time the patch version is updated.  If we leave it as it is, we only have to do a migration every time the minor version gets updated.  So unless you expect breaking changes in the ABI to occur in patch updates, we should leave this.

It doesn't look like there should be room for concern to me:
https://abi-laboratory.pro/?view=timeline&l=gdal

xylar (approved):

rouault:

PR #992, opened 2024-10-24
Update to GDAL 3.10.0, and add libgdal-avif and libgdal-heif

rouault:
Checklist
* [X] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [X] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [X] Ensured the license file is being packaged.

PR #1046, opened 2025-01-23
Remove libheif runtime dependency from libgdal-core

himikof:
This dependency was introduced in https://github.com/conda-forge/gdal-feedstock/pull/992, and is clearly incorrect.

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -11,7 +11,7 @@ source:
     - 000_cmake.patch  # [osx]
 
 build:
-  number: 2
+  number: 3
   skip_compile_pyc:
     - share/bash-completion/completions/*.py
   ignore_run_exports_from:
@@ -20,6 +20,7 @@ build:
     - openjpeg
     - tiledb
     - libnetcdf
+    - libheif
   run_exports:
     # no idea, going with minor pin
     - {{ pin_subpackage('libgdal-core', max_pin='x.x') }}


akrherz:
Looking at most recent main branch build logs I find

```
2025-01-23T05:35:44.1273074Z WARNING (libgdal-core): run-exports library package conda-forge/linux-64::libheif==1.19.5=gpl_hc21c24c_100 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
2025-01-23T05:35:44.1273795Z WARNING (libgdal-core): run-exports library package conda-forge/linux-64::libuuid==2.38.1=h0b41bf4_0 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
```

Should `libuuid` be added to `ignore_run_exports_from` as well?

himikof:
I have no idea why would `libuuid` be needed at all. Quick googling tells that this once was a dependency of `libdap`, which in turn was needed by DODS/OpenDAP drivers. Which were deprecated and removed in 3.5.0: https://github.com/OSGeo/gdal/pull/5666.

So maybe this dependency should just be dropped from the environment altogether?

isuruf:
Yes, you can drop `libuuid`

rouault:
The dependency is needed for getting deferred loaded driver metadata correct. Only headers are used. libgdal-core is not linked against it. See https://github.com/OSGeo/gdal/issues/12245 for an extensive discussion about that tricky topic

PR #1087, opened 2025-05-13
Add JPEGXL-in-TIFF support and JPEGXL driver

### LINKED ISSUE #752: request: add jpegxl as a dependency
### Comment:

In contrast to #722, I'd like to request the addition of a new dependency of JPEG-XL (https://github.com/libjxl/libjxl). 

This would enable the jpegxl driver for GDAL and many neat compression options that brings to COGs.

rouault:
For the former we have to use internal libtiff and geotiff. I'm not totally sure if it is OK w.r.t conda-forge packaging policies ? Technicially it is safe, since the symbols of those vendored libraries are prefixed with GDAL_ to avoid any clash if  an application/library linking against GDAL is also linked against libtiff

Fixes #752

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -11,7 +11,7 @@ source:
     - 000_cmake.patch  # [osx]
 
 build:
-  number: 1
+  number: 2
   skip_compile_pyc:
     - share/bash-completion/completions/*.py
   ignore_run_exports_from:
@@ -39,7 +39,8 @@ requirements:
     - blosc
     - expat
     - geos
-    - geotiff
+    # We use internal libtiff and libgeotiff for JXL-in-TIFF support
+    # - geotiff
     - giflib  # [not win]
     - libjpeg-turbo
     - json-c  # [not win]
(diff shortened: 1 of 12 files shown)


akrherz (approved):
This seems reasonable to me, but would hopefully get somebody more knowledgeable on conda-forge to also sign-off..

recipe/meta.yaml
```diff
@@ -11,7 +11,7 @@ source:
     - 000_cmake.patch  # [osx]
 
 build:
-  number: 0
+  number: 1
```
akrherz, line 14: Need the build number bumped another one now.

rouault:

recipe/meta.yaml
rouault: done

akrherz:
@xylar Could you review this as well?

xylar:
Thanks for the ping.  @rouault and @akrherz, my understanding is that vendoring these libraries is fine as long as they symbols are unique and you include the licenses.  So let's add the license names and license files here:
https://github.com/rouault/gdal-feedstock/blob/jxl/recipe/meta.yaml#L764-L765
I couldn't find anything in the existing `LICENSE.TXT` to cover either of the vendored libraries.

rouault:
> So let's add the license names and license files here:
> https://github.com/rouault/gdal-feedstock/blob/jxl/recipe/meta.yaml#L764-L765

done

Issue #1119, opened 2025-09-10
GPL compliancy regarding libarchive that uses lzo


deltamarnix:
### Comment:

Dear feedstock maintainers,

I'm looking into packaging my conda environment into an installer that can be sent to end users for redistribution. The installer will contain libgdal-core. The self-developed package that uses libgdal-core is licensed under MIT.

But since I'm redistributing the package, I foresee a problem. libgdal-core uses libarchive (licensed under BSD-2), which in turn uses lzo (licenced under GPL).

It seems like, if I redistribute the conda forge version of libgdal-core, I am also redistributing lzo. And that would be in conflict with the GPL license.

libarchive has a build variant that can be found in [conda_build_config.yaml](https://github.com/conda-forge/libarchive-feedstock/blob/main/recipe/conda_build_config.yaml). Setting the license family to `lgpl` will remove lzo support, making the libarchive package LGPL again.

To my understanding, this flag has to be set in the conda recipe that is using the underlying library. Would it be possible to set the flag to lgpl, so that lzo is no longer supported?

rouault:
> Would it be possible to set the flag to lgpl, so that lzo is no longer supported?

I'm no familiar with that flag, but GDAL only uses the 7zip and rar components of libarchive, which I don't think require lzo

deltamarnix:
I'm trying to create a new PR where I will use libarchive without lzo. Can I simply change the usage in the meta to: `    - libarchive * lgpl*`? Since lzo isn't used anyway.

Or would it be better to create a gpl / lgpl split for gdal to ensure that people can choose which version of libarchive they want to use?

PR #1130, opened 2025-10-03
Use lgpl versions of libarchive and libspatialite

### LINKED ISSUE #1118: GPL compliancy regarding libspatialite that uses librttopo
### Comment:

Dear feedstock maintainers,

I'm looking into packaging my conda environment into an installer that can be sent to end users for redistribution. The installer will contain libgdal-core. The self-developed package that uses libgdal-core is licensed under MIT.

But since I'm redistributing the package, I foresee a problem. libgdal-core uses libspatialite (tri-licensed under MPL, ,GPL, LGPL), which in turn uses librttopo (licenced under GPL).

It seems like, if I redistribute the conda forge version of libgdal-core, I am also redistributing librttopo. And that would be in conflict with the GPL license.

Since libspatialite is under a cmake flag, it looks like it is being linked and built. Is it possible to move libspatialite to a plugin structure, so that it is no longer part of libgdal-core? Otherwise I will have to reconsider my way of distribution and build gdal myself, which would be a pity.

deltamarnix:
LZO and RTTOPO aren't really used by GDAL, so removing them from the build makes it possible to download and redistribute GDAL without trespassing the GPL license of underlying packages.

Fixes #1118 #1119

Checklist
* [x] Reset the build number to `0` (if the version changed)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -11,7 +11,7 @@ source:
     - 000_cmake.patch  # [osx]
 
 build:
-  number: 4
+  number: 5
   skip_compile_pyc:
     - share/bash-completion/completions/*.py
   ignore_run_exports_from:
@@ -45,7 +45,7 @@ requirements:
     - libjpeg-turbo
     - json-c  # [not win]
     - lerc
-    - libarchive
+    - libarchive * lgpl*
     - libcurl
     - libdeflate
     - libiconv
@@ -53,7 +53,7 @@ requirements:
     - libkml-devel
     - liblzma-devel
     - libpng
-    - libspatialite
+    - libspatialite * lgpl*
     # We use internal libtiff and libgeotiff for JXL-in-TIFF support
     # - libtiff
     - libwebp-base


recipe/meta.yaml
```diff
@@ -45,15 +45,15 @@ requirements:
...
+    - libarchive * lgpl*
     - libcurl
     - libdeflate
     - libiconv
     - libjxl
     - libkml-devel
     - liblzma-devel
     - libpng
-    - libspatialite
+    - libspatialite * lgpl*
```
xylar, line 56: I just want to make sure about this.  Wouldn't this mean that folks can't install `(lib)gdal` and `libspatialite` with rttopo support in the same environment?  Even if `gdal` itself doesn't use these features, we may need a variant that supports gpl so we don't prevent this.

@conda-forge/gdal, what to other maintainers think?

deltamarnix: I'm not well educated in the GDAL source code, but what I could imagine is that libspatialite is built without RTTOPO, meaning that it will not search for it, will not link to it, and will interact with it.
It also means that GDAL linked to libspatialite would not be able to use any RTTOPO features. As I understood from #1118 , this shouldn't really be necessary.

If anyone would still add rttopo from the librttopo-feedstock, I don't think that should be a problem, and users would still be able to interact with rttopo directly.

I would also like to know from the gdal maintainers if it is indeed true that rtopo features are not used from libspatialite. But if I only look at it from the possibility to still download libspatialite separately, I would say that it would still be possible this way.

recipe/meta.yaml
```diff
@@ -45,15 +45,15 @@ requirements:
     - libjpeg-turbo
     - json-c  # [not win]
     - lerc
-    - libarchive
+    - libarchive * lgpl*
```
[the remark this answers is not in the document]
deltamarnix: Would I also need to do that for libspatialite?

deltamarnix:

recipe/meta.yaml
deltamarnix: I agree that that would be painful. Also, it might mean that any downstream packages that are using gdal also require gpl/lgpl variants, causing a whole chain of new build variants on conda forge. I wouldn't want to be the causer of that. I hope that this use case is non-existent, so let's see what others say. 🤞

PR #1203, opened 2026-05-08
remove setuptools from staging build

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)

### DIFF
--- a/recipe/recipe.yaml
+++ b/recipe/recipe.yaml
@@ -2,7 +2,7 @@ schema_version: 1
 
 context:
   version: "3.13.0"
-  build_num: 3
+  build_num: 1
 
 recipe:
   name: libgdal-split
--- a/recipe/recipe.yaml
+++ b/recipe/recipe.yaml
@@ -2,7 +2,7 @@ schema_version: 1
 
 context:
   version: "3.12.4"
-  build_num: 2
+  build_num: 3
 
 recipe:
   name: libgdal-split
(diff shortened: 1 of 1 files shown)


rouault:
I presume this is to fix https://github.com/prefix-dev/rattler-build/issues/2481 ? Looking at the logs, this does seem to do the trick! no more traces of ``_distutils_hack``

isuruf:
Yes, it is

rouault (commented):
the build number needs to be bumped

akrherz:
Ah, this needs rebased as gdal 3.13 is a thing now :)

recipe/recipe.yaml
```diff
@@ -2,7 +2,7 @@ schema_version: 1
 
 context:
   version: "3.13.0"
-  build_num: 2
+  build_num: 3
```
akrherz: There is build number troubles with the 3.12.x branch, but for main branch, this should be `1` I believe.

akrherz (approved):

recipe/recipe.yaml
xylar: The issue is that `build_num` is not being used and `build` is still set to zero below. I will fix it.

xylar: Oh, sorry, I see that it's fixed here.  I was looking at it being messed up on the main branch.  sorry for the noise!

PR #1236, opened 2026-08-30
Enable Windows ARM64 GDAL with full Windows feature coverage

ClayWarren:
Enable native Windows ARM64 builds with the same features, plugin outputs, and Python bindings as existing Windows builds. Removes the earlier ARM64-only exclusions and rerenders the full matrix.

Draft pending publication of the required Windows ARM64 dependencies. Recipe lint and rendering pass; full native build and runtime validation remain pending. Rerender also removes Python 3.10 variants under current conda-forge pinning.

Prepared with assistance from OpenAI Codex.


hobu:
This shuts off a bunch of things for win_arm64 that shouldn't be deactivated. The dependencies that are needed should instead have win_arm64 packages built and pushed.

xylar:
I agree.  Unless this is blocking migration of its own dependencies (e.g. via some sort of circular dependency), it would be much less confusing to just wait for all the upstream dependencies to be in place.

